Issue #23, opened 2020-10-14
Forwarding mode cannot work reliably as currently defined


DavidSchinazi:
Forwarding mode works by having the proxy demultiplex incoming packets by looking at the Destination Connection ID field in short headers: if it's a registered "Server Connection ID", it gets forwarded to the target; otherwise, it gets processed as a QUIC packet intended for the proxy itself.

Unfortunately, Server Connection IDs are not under the control of the proxy, they're under the control of the target(s). So that means that there can easily be conflicts. The fact that short header packets do not carry the connection ID length complicates this further: any connection being a prefix of another one also constitutes a conflict. In the worse case scenario, a single target using zero-length connection IDs would conflict with all other Server Connection IDs.

In #17 and #18, a proposed solution is to reject CONNECT-QUIC requests whose Server Connection IDs conflict with existing mappings. This would mean that this protocol would be completely unable to handle a set of targets, which would be unfortunate.

However, we could solve this issue by making the forwarding mode negotiable. Here's a straw man: the client adds a `Forwarding-Mode: 0/1` header to its request to tell the proxy whether it wants forwarding mode for this request or not, and the proxy replies with the same header to indicate whether forwarding mode is allowed for this request. That way, if the server detects a conflict, it gracefully falls back to tunneled mode and everything still works. (Bonus 1: this also allows us to support proxies behind load balancers, they just need to disable forwarding) (Bonus 2: this also allows us to support the [MASQUE Obfuscation](https://tools.ietf.org/html/draft-schinazi-masque-obfuscation-03) use-case by letting the client disable forwarding)

tfpauly:
Yeah, I was expecting that a forwarding mapping failing would mean we'd have to fall back to tunneled mode anyhow. I'll specify this more clearly, as it is the intent.

tfpauly:
Addressed by #24

Issue #22, opened 2020-10-14
Is the concept of a "Client Connection ID" needed?


DavidSchinazi:
The document mentions the concept of a "Client Connection ID", but I'm not sure they serve any purpose.

For example:
> Clients SHOULD establish mappings for any Client Connection ID values it provides to the destination target. Failure to do so will prevent the target from initiating connection migration probes along new paths.

What is the Client Connection ID mapped *to*? (see #21 about clarifying that)

If you create one UDP socket per CONNECT-QUIC request, then you don't need the concept of a Client Connection ID. The server knows how to route packets simply based on the UDP socket they came in on.

The draft mentions validating that a Client Connection ID has been properly registered before allowing its use, but I don't see what benefit that provides.

I feel like this concept might have been inherited from [the original MASQUE QUIC Proxying](https://tools.ietf.org/html/draft-schinazi-masque-protocol-01#section-3.3) but there it was needed to allow the MASQUE server to only use one UDP port, which isn't a requirement here. Am I missing something?

tfpauly:
Addressed by #24

Issue #21, opened 2020-10-14
Mapping details don't seem right


DavidSchinazi:
The only details about how mappings work seem to be this part of the "Proxy Response Behavior" section:
> Creating a mapping entry for the QUIC Connection ID in the given direction (client or target server) associated with the client's IP address and UDP port. For any non-zero-length Client Connection ID, the Connection ID MUST be unique across all other clients.

That's not right. You need at least the following mappings:
- from flow ID to target IP/port (for tunneling)
- from Server Connection ID to target IP/port (for forwarding)
- from target IP/port to (QUIC connection, flow ID) (for tunneling)
- from target IP/port to client IP/port (for forwarding)

tfpauly:
Addressed by #24

Issue #20, opened 2020-10-14
Reusing DATAGRAM Flow IDs


DavidSchinazi:
In the Example section, the spec reuses the same Flow ID for two separate CONNECT-QUIC requests. That seems like a reasonable choice to me, but the spec should explicitly call out that this is allowed, and that the proxy should validate that upon receiving a new CONNECT-UDP request that reuses a flow ID, the authority must match the authority of the first CONNECT-QUIC request that used the same flow ID.

tfpauly:
Agreed, good to clarify.

tfpauly:
Agreed, good to clarify.

tfpauly:
Addressed by #24

PR #24, opened 2020-10-15
Define proxy mapping state more rigorously

tfpauly:
Addressing #23, #22, #21, #20

### DIFF
--- a/draft-pauly-connect-quic.md
+++ b/draft-pauly-connect-quic.md
@@ -59,12 +59,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
-# Proxying using QUIC Connection IDs
+# Required Proxy State {#mappings}
 
-The method described in this document proxies flows at the granularity of QUIC
-Connection IDs, which are exposed in QUIC packets and can be used for
-identifying and routing the packets. Each QUIC Connection ID represents one direction
-of QUIC packets, with the Connection ID being owned and defined by the receiver of the packets.
+In the methods defined in this document, the proxy is aware of the QUIC Connection IDs
+being used by proxied connections, along with the sockets (UDP addresses and ports)
+used to communicate with the client and the target. Tracking Connection IDs in this way
+allows the proxy to reuse sockets for multiple connections and support the forwarding
+mode of proxying.
+
+A QUIC Connection ID identifies the receiver of a packet, and is defined by the receiver.
 
 A Connection ID that is defined by the client of HTTP/3 proxy, and is used
 to route packets from the target server to the client when it is a Destination Connection ID,
@@ -78,6 +81,62 @@ QUIC packets can be either tunnelled within an HTTP/3 proxy connection using
 QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
 connection on the same set of IP addresses and UDP ports. CONNECT-QUIC allows
 clients to specify either form of transport.
+
+In order to correctly route QUIC packets in both tunnelled and forwarded modes, the proxy
+needs to maintain mappings between several items:
+
+- Datagram flow, which is a flow of QUIC DATAGRAM frames specific to a single client QUIC connection to the proxy.
+- Client-facing socket, which is the set of UDP addresses and ports used to communicate between the client and the proxy.
+- Server-facing socket, which is the set of UDP addresses and ports used to communicate between the proxy and the target.
+- Client Connection ID, which is a QUIC Connection ID used to route traffic to a client.
+- Server Connection ID, which is a QUIC Connection ID used to route traffic to a target.
+
+There are three required mappings, described below.
+
+## Datagram Flow Mapping
+
+Each datagram flow MUST be mapped to a single server-facing socket. This datagram flow can be identified by datagram flow ID
+within a unique QUIC connection between the client and the proxy.
+
+~~~
+Datagram flow => Server-facing socket
+~~~
+
+Multiple datagram flows can map to the same server-facing socket, but a single datagram flow cannot be mapped to multiple
+server-facing sockets.
+
+This mapping guarantees that any QUIC packet sent from the client to the proxy in tunnelled mode can be sent to the correct
+target.
+
+## Server Connection ID Mapping
+
+Each pair of Server Connection ID and client-facing socket MUST map to a single server-facing socket.
+
+~~~
+(Client-facing socket + Server Connection ID) => Server-facing socket
+~~~
+
+Multiple pairs of Connection IDs and sockets can map to the same server-facing socket.
+
+This mapping guarantees that any QUIC packet sent from the client to the proxy in forwarded mode can be sent to the correct
+target.
+
+## Client Connection ID Mappings
+
+Each pair of Client Connection ID and server-facing socket MUST map to a single client-facing socket.
+Additionally, that same pair MUST map to a single datagram flow.
+
+~~~
+(Server-facing socket + Client Connection ID) => Client-facing socket
+(Server-facing socket + Client Connection ID) => Datagram flow
+~~~
+
+Multiple pairs of Connection IDs and sockets can map to the same client-facing socket or datagram flow.
+
+These mapping guarantee that any QUIC packet sent from a target to the proxy in either tunnelled or forwarded
+mode can be sent to the correct client. Note that this mapping becomes trivial if the proxy always opens a new
+server-facing socket for ever proxied QUIC connection. The mapping is critical for any case where server-facing
+sockets are shared or reused.
 
 # The CONNECT-QUIC Method {#connect-quic-method}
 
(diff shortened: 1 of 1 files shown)


draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+
+# Proxy Response Behavior
+
+Upon receipt of a CONNECT-QUIC request, the proxy validates the request,
+tries to establish the appropriate mappings described in {{mappings}}, and
+establish a new server-facing socket if necessary.
+
+The proxy MUST validate that the request includes either the Client-Connection-Id or
+the Server-Connection-Id header, along with a Datagram-Flow-Id header and an
+authority pseudo-header. If any of these is missing, the proxy MUST reject the
```
LPardue, line 293: See earlier comment about more terse terms. We could then tighten this up to say soemthing "The proxy MUST validate that the CONNECT-QUIC request is either type X or Y..."

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+## Removing Mapping State
 
 Each CONNECT-QUIC request consumes one bidirectional HTTP/3 stream. For any stream
-on which the proxy has sent a response indicating success, the mapping for forwarding a
-Connection ID lasts as long as the stream is open.
+on which the proxy has sent a response indicating success, any mappings for the request
+last as long as the stream is open.
 
 A client that no longer wants a given Connection ID to be forwarded by the proxy, for either
 direction, MUST cancel its CONNECT-QUIC HTTP/3 request {{!I-D.ietf-quic-http}}.
```
LPardue, line 340: This section reads a little weird. I'll make a spin off issueo work on improvements for talking about request stream lifecycle; it should also cover the similar stuff described in H3's CONNECT section like what frames are valid.

LPardue (commented):
This is super helpful. Thanks. 

I've got suggestions for nits. But also there's some things I think could be improvements.

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
 connection on the same set of IP addresses and UDP ports. CONNECT-QUIC allows
 clients to specify either form of transport.
 
+In order to correctly route QUIC packets in both tunnelled and forwarded modes, the proxy
+needs to maintain mappings between several items:
+
+- Datagram flow, which is a flow of QUIC DATAGRAM frames specific to a single client QUIC connection to the proxy.
```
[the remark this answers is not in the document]
tfpauly:
```suggestion
- Datagram flow, which is a flow of HTTP/3 DATAGRAMs specific to a single client QUIC connection to the proxy.
```

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
...
+within a unique QUIC connection between the client and the proxy.
+
+~~~
+Datagram flow => Server-facing socket
+~~~
+
+Multiple datagram flows can map to the same server-facing socket, but a single datagram flow cannot be mapped to multiple
+server-facing sockets.
+
+This mapping guarantees that any QUIC packet sent from the client to the proxy in tunnelled mode can be sent to the correct
```
DavidSchinazi, line 108: Well technically the client could send a packet with a flow ID that's unknown to the server

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
 connection on the same set of IP addresses and UDP ports. CONNECT-QUIC allows
 clients to specify either form of transport.
 
+In order to correctly route QUIC packets in both tunnelled and forwarded modes, the proxy
+needs to maintain mappings between several items:
+
+- Datagram flow, which is a flow of HTTP/3 DATAGRAMs specific to a single client QUIC connection to the proxy.
+- Client-facing socket, which is the set of UDP addresses and ports used to communicate between the client and the proxy.
```
DavidSchinazi, line 89: Our implementation doesn't use connected sockets on the server - each QUIC server has a single client-facing non-connected UDP socket. Maybe just mentioning somewhere that this spec uses "socket" as a way to refer to a 4-tuple and that it can be implemented a variety of ways

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
...
+This mapping guarantees that any QUIC packet sent from the client to the proxy in forwarded mode can be sent to the correct
+target.
+
+## Client Connection ID Mappings
+
+Each pair of Client Connection ID and server-facing socket MUST map to a single client-facing socket.
+Additionally, that same pair MUST map to a single datagram flow.
+
+~~~
+(Server-facing socket + Client Connection ID) => Client-facing socket
```
DavidSchinazi, line 130: I don't think that's right, the mapping should be to a QUIC connection, not a client-facing socket - because multiple QUIC connections can share a client-facing socket

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
...
+(Server-facing socket + Client Connection ID) => Datagram flow
+~~~
+
+Multiple pairs of Connection IDs and sockets can map to the same client-facing socket or datagram flow.
+
+These mapping guarantee that any QUIC packet sent from a target to the proxy in either tunnelled or forwarded
+mode can be sent to the correct client. Note that this mapping becomes trivial if the proxy always opens a new
+server-facing socket for every proxied QUIC connection. The mapping is critical for any case where server-facing
```
DavidSchinazi, line 138: I'm not sure we've defined the term "proxied QUIC connection" anywhere. I ask because that's a concept that the proxy doesn't really have access to - the proxy can't distinguish between "two CIDs belonging to one connection" vs "two connections" if only short headers are involved

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
-QUIC packets via the proxy, it sends a new CONNECT-QUIC request.
+A clients sends new CONNECT-QUIC requests when it wants to start
+a new QUIC connection to a target, when it has received a new
+Server Connection ID for the target, and before it advertises a new Client
+Connection ID to the target.
 
-Clients can choose to send QUIC packets to the proxy either tunnelled within
-DATAGRAM frames, or sent directly to the proxy's IP address and port.
+Each request MUST contain a Datagram-Flow-Id header and an authority
+psuedo-header identifying the target. All requests for the same QUIC
```
DavidSchinazi, line 196:
```suggestion
pseudo-header identifying the target. All requests for the same QUIC
```

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+Connection between a client and a target SHOULD contain the same Datagram-Flow-Id
```
DavidSchinazi, line 197: Why is this a SHOULD and not a MUST? It's required for correctness

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+and authority. Any mismatch in these would cause the proxy to treat the requests
```
DavidSchinazi, line 198: We should address the discrepancy between authority and server-facing 4-tuple, they're not quite the same because DNS can return multiple results so we need to mention that the proxy has to save the authority in the mappings and reuse the DNS result from the first time around. (I'm also realizing the proxy really shouldn't perform happy eyeballs here because that would make things way more complicated)

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
-CONNECT-QUIC request has been sent, even in the same QUIC packet to the proxy.
-That is, the QUIC packet sent from the client to the proxy can contain a STREAM
-frame with the CONNECT-QUIC request, as well as a DATAGRAM frame that contains
-a tunnelled QUIC packet to forward. This is particularly useful for reducing round trips.
+## New Proxied Connection Setup
+
+The first time that a client uses a proxy for a given QUIC connection, it selects a new datagram
+flow ID with an even-numbered value {{!I-D.schinazi-quic-h3-datagram}}.
+
+The first request the clients makes MUST contain the authority psuedo-header and the
```
DavidSchinazi, line 212:
```suggestion
The first request the clients makes MUST contain the authority pseudo-header and the
```

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+first CONNECT-QUIC request for the datagram flow ID has been sent, even in the same QUIC
+packet to the proxy. That is, the QUIC packet sent from the client to the proxy can contain a
+STREAM frame containing the CONNECT-QUIC request, as well as a DATAGRAM frame that contains
+a tunnelled QUIC packet to send to the target. This is particularly useful for reducing round trips
+on connection setup.
+
+Since clients are always aware whether or not they are using a QUIC proxy, clients are
+expected to cooperate with proxies in selecting Client Connection IDs. A proxy
+detects a conflict when it is not able to create a unique mapping using the Client Connection ID. 
+It can reject requests that would cause a conflict and indicate this to the client by replying with a
```
DavidSchinazi, line 227: We should define the term of "conflict" before using it. It particular, one CID being a prefix of another is a "conflict" for the purposes of this draft.

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
...
+
+~~~
+Datagram flow => Server-facing socket
+~~~
+
+Multiple datagram flows can map to the same server-facing socket, but a single datagram flow cannot be mapped to multiple
+server-facing sockets.
+
+This mapping guarantees that any QUIC packet sent from the client to the proxy in tunnelled mode can be sent to the correct
+target.
```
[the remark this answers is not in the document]
tfpauly: I just tried arranging it the other way, but I think it was less clear. It flows right now as:
- Text description of the mapping requirement
- Diagram of the mapping
- Explanation for what the mapping means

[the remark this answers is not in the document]
LPardue: Yeah that works for me.

draft-pauly-connect-quic.md
tfpauly: Yeah, I think I may leave that for a later editorial cleanup. I am hesitant to add too much jargon yet.

draft-pauly-connect-quic.md
tfpauly: Sounds good, new issue welcome (and text welcome too!)

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+## Sending With Forwarded Mode
+
+Once the client has learned the target server's Connection ID, such as in the response
+to a QUIC Initial packet, it can send a request containing the Server-Connection-Id
+header. The client MUST wait for a successful 200 (OK) response before using forwarded
+mode. Prior to receiving the server response, the client MUST send short header packets
+tunnelled in DATAGRAM frames. The client MAY also choose to tunnel some short header
+packets even after receiving the successful response.
+
+If the client receives an error, such as a 409 (Conflict) response, it MUST NOT forward
```
DavidSchinazi: This is ambiguous: the server could decide to reject the request because it doesn't like the authority (e.g. if the client sends "localhost" as the authority, the proxy really ought to refuse to proxy there). Another example is DNS failure. I think that we need a way to convey the scenario where "the proxy agrees to proxy to this authority and has opened a target-facing socket, but cannot use forwarding mode due to a conflict".

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+packets to the requested Server Connection ID, but only use tunnelled mode.
```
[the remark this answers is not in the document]
tfpauly: Fixed this up!

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+QUIC long header packets MUST NOT be forwarded. These packets can only be tunnelled within
+DATAGRAM frames to avoid exposing unnecessary connection metadata.
+
+When forwarding, the client sends a QUIC packet with the target server's Connection ID
+in the QUIC short header, using the same socket between client and proxy that was used
+for the main QUIC connection between client and proxy.
+
+## Receiving With Forwarded Mode
+
+Once a Client Connection ID has been accepted by the proxy, the client MUST be prepared to
```
DavidSchinazi, line 281: As discussed in #23, it would be great to allow clients to disable forwarding mode

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+
+Upon receipt of a CONNECT-QUIC request, the proxy validates the request,
+tries to establish the appropriate mappings described in {{mappings}}, and
+establish a new server-facing socket if necessary.
+
+The proxy MUST validate that the request includes either the Client-Connection-Id or
+the Server-Connection-Id header, along with a Datagram-Flow-Id header and an
+authority pseudo-header. If any of these is missing, the proxy MUST reject the
+request with a 400 (Bad Request) reponse. The proxy also MUST reject the request
+if the requested datagram flow ID has already been used on that client QUIC connection
```
DavidSchinazi, line 296:
```suggestion
if the requested datagram flow ID has already been used on that client-to-proxy QUIC connection
```

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+
+The proxy then determines the server-facing socket to associate with the client's
+datagram flow. This UDP socket may already be open (from a previous request from this
+client, or another). If the socket is not already created, the proxy creates a new one.
+Proxies can choose to reuse server-facing sockets across multiple datagram flows, or
+have a unique server-facing socket for every datagram flow.
+
+If the request includes a Client-Connection-Id header, the proxy is receiving a request
+to be able to route traffic back to the client using that Connection ID. If the pair of this
+Client Connection ID and the selected server-facing socket is unique, the proxy creates
```
DavidSchinazi, line 307: I would phrase this as "avoiding conflict" instead of "unique" because conflicts here aren't only about equality

DavidSchinazi (changes requested):
Thanks for writing this, it really addresses a lot of things! Sorry about the mountain of comments, many are editorial but some are design issues that we'd need to address if we want to resolve the 4 issues

draft-pauly-connect-quic.md
tfpauly: Yes, that's correct—and the QUIC connection implies the client-facing socket and datagram flow ID. Can switch the language.

draft-pauly-connect-quic.md
tfpauly: So, my thought here was that a client _could_ decide to be sneaky and represent what is one connection between the client and the target as two subsequent flows to proxy, which could make it "migrate". It'd be some paranoid privacy thing, but I could see a reason it would be illegal as long as the client followed the rest of the rules.

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+Client Connection ID, the client MUST select a new Connection ID before sending
+a new request, and generate a new packet. For example, if a client is sending a
+QUIC Initial packet and chooses a Connection ID that is too short or hits a conflict
+with an existing mapping to the same target server, it will need to generate a new
+QUIC Initial.
+
+## Adding new Client Connection IDs
+
+A client can add new Connection IDs to a proxied QUIC connection by sending
+a NEW_CONNECTION_ID frame.
```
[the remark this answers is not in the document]
tfpauly:
```suggestion
A client can add new Connection IDs to a proxied QUIC connection by sending
a NEW_CONNECTION_ID frame to the target.
```

draft-pauly-connect-quic.md
tfpauly: So my thought here is that disabling forwarding is essentially just using CONNECT-UDP (which is fine). My question is if that is a separate method, like CONNECT-UDP, or if it is a CONNECT-QUIC request that has a flow ID and authority but no connection ID.

draft-pauly-connect-quic.md
tfpauly:
```suggestion
if the requested datagram flow ID has already been used on that client <-> proxy QUIC connection
```

draft-pauly-connect-quic.md
tfpauly:
```suggestion
Client Connection ID and the selected server-facing socket does not create a conflict, the proxy creates
```

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,167 @@ an Integer. Its ABNF is:
...
+ID can to be sent to the client. The proxy MUST use tunnelled mode (DATAGRAM frames) on
+the correct datagram flow for any long header packets. The proxy SHOULD forward directly to
+the client for any matching short header packets, but MAY tunnel them in DATAGRAM frames.
+If the pair is not unique, or the proxy chooses not to support zero-length Client Connection IDs,
+the proxy responds with a 409 (Conflict) response. If this occurs on the first request for a given datagram flow,
+the proxy removes any mapping for that datagram flow.
+
+If the request includes a Server-Connection-Id header, the proxy is receiving a request
+to allow the client to forward packets to the target. If the pair of this Server Connection ID
+and the client-facing socket on which the request was received is unique, the proxy
```
tfpauly, line 319:
```suggestion
and the client-facing socket on which the request was received does not create a conflict, the proxy
```

draft-pauly-connect-quic.md
```diff
@@ -59,12 +59,15 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
-# Proxying using QUIC Connection IDs
+# Required Proxy State {#mappings}
 
-The method described in this document proxies flows at the granularity of QUIC
-Connection IDs, which are exposed in QUIC packets and can be used for
-identifying and routing the packets. Each QUIC Connection ID represents one direction
-of QUIC packets, with the Connection ID being owned and defined by the receiver of the packets.
+In the methods defined in this document, the proxy is aware of the QUIC Connection IDs
+being used by proxied connections, along with the sockets (UDP addresses and ports)
```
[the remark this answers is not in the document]
tfpauly: Clarified

draft-pauly-connect-quic.md
```diff
@@ -79,6 +82,62 @@ QUIC DATAGRAM frames, or be forwarded directly alongside an HTTP/3 proxy
...
+- Client-facing socket, which is the set of UDP addresses and ports used to communicate between the client and the proxy.
+- Server-facing socket, which is the set of UDP addresses and ports used to communicate between the proxy and the target.
+- Client Connection ID, which is a QUIC Connection ID used to route traffic to a client.
+- Server Connection ID, which is a QUIC Connection ID used to route traffic to a target.
+
+There are three required mappings, described below.
+
+## Datagram Flow Mapping
+
+Each datagram flow MUST be mapped to a single server-facing socket. This datagram flow can be identified by datagram flow ID
```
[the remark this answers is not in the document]
tfpauly: Updated the mapping to that!

draft-pauly-connect-quic.md
tfpauly: Clarified the text to be about this particular flow ID

draft-pauly-connect-quic.md
tfpauly: Updated the mappings. I think there are still two mappings — one to the QUIC Connection + Datagram Flow ID, and one to the client-facing socket (for forwarding).

draft-pauly-connect-quic.md
tfpauly: Added a note about what the term "socket" means for this document.

draft-pauly-connect-quic.md
tfpauly: Changed this to refer to the datagram flow ID, which is what it actually knows.

draft-pauly-connect-quic.md
tfpauly: Added a paragraph about this in the proxy behavior section, since it really belongs there (the client doesn't really care)

draft-pauly-connect-quic.md
tfpauly: Added a section early on (in state requirements) about what conflicts mean.

draft-pauly-connect-quic.md
tfpauly: Requests that include Server-Connection-Id are only about opening up forwarding paths, not creating new upstream connections. Thus, I think the correct response here is to not forward, whatever the particular error code; assuming that you already were able to open up a tunneling path with the earlier request.

draft-pauly-connect-quic.md
DavidSchinazi: Good point, agreed.

draft-pauly-connect-quic.md
DavidSchinazi: Thanks, I like that.

draft-pauly-connect-quic.md
DavidSchinazi: That makes sense for the flow ID, but not for the authority - perhaps we say MUST be sent to the same authority and SHOULD be sent to the same Datagram-Flow-Id?

draft-pauly-connect-quic.md
```diff
@@ -126,93 +202,169 @@ an Integer. Its ABNF is:
...
+if the requested datagram flow ID has already been used on that client <-> proxy QUIC connection
+with a different requested authority.
+
+The proxy then determines the server-facing socket to associate with the client's
+datagram flow. This UDP socket may already be open (from a previous request from this
+client, or another). If the socket is not already created, the proxy creates a new one.
+Proxies can choose to reuse server-facing sockets across multiple datagram flows, or
+have a unique server-facing socket for every datagram flow.
+
+If a proxy reuses server-facing sockets, it SHOULD store which authorities (server names)
```
DavidSchinazi, line 320:
```suggestion
If a proxy reuses server-facing sockets, it SHOULD store which authorities (which could be a domain name or IP address literal)
```

draft-pauly-connect-quic.md
```diff
@@ -126,93 +202,169 @@ an Integer. Its ABNF is:
...
+are being accessed over a particular server-facing socket so it can avoid performing a
+new DNS query and potentially choosing a different server IP address.
```
DavidSchinazi, line 322:
```suggestion
new DNS query and potentially choosing a different server IP address which could map to a different server.
```

draft-pauly-connect-quic.md
DavidSchinazi: Agreed, that's a better spot for it - I'll create new comments over there.

draft-pauly-connect-quic.md
```diff
@@ -126,93 +202,169 @@ an Integer. Its ABNF is:
...
+
```
DavidSchinazi, line 323: I'd suggest adding some text about how the server needs to resolve domain name authorities to IP addresses in the first place, we have text for reusing sockets but this might warrant its own paragraph

draft-pauly-connect-quic.md
DavidSchinazi: New text LGTM

draft-pauly-connect-quic.md
DavidSchinazi: I think this complicates the design - I'd rather think about any request being able to open up any mapping or socket - and I'd rather not use 4xx to indicate partial success - there are probably implementations out there that close the stream if the response isn't 2xx.

draft-pauly-connect-quic.md
DavidSchinazi: I think there's value in using CONNECT-QUIC without forwarding: one is allowing the proxy to only need one socket, another is if we decide to later add short header connection ID compression to CONNECT-QUIC

draft-pauly-connect-quic.md
tfpauly: Added some text on this above!

draft-pauly-connect-quic.md
```diff
@@ -126,93 +185,166 @@ an Integer. Its ABNF is:
...
+returns a 409 (Conflict) error. Any DATAGRAM frames that are sent in a separate
+QUIC packet from the STREAM frame that contains the CONNECT-QUIC request might
+also be dropped in the case that the packet arrives at the proxy before the packet
+containing the STREAM frame.
+
+If the server rejects the first request that uses a specific datagram flow ID, the client
+MUST retire that datagram flow ID. If the rejection indicated a conflict due to the
+Client Connection ID, the client MUST select a new Connection ID before sending
+a new request, and generate a new packet. For example, if a client is sending a
+QUIC Initial packet and chooses a Connection ID that is too short or hits a conflict
```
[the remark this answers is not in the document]
tfpauly: Removed that bit—it's now explained in the "conflicts" section.

draft-pauly-connect-quic.md
tfpauly: Okay, updated text to allow either side to opt out of forwarding!

draft-pauly-connect-quic.md
tfpauly: Updated the text to clarify this. The stream will indeed be closed and failed with any 4xx response (that just doesn't invalidate a separate stream that requested non-forwarding proxying to the same authority).

draft-pauly-connect-quic.md
```diff
@@ -135,93 +214,189 @@ an Integer. Its ABNF is:
   Datagram-Flow-Id = sf-integer
 ~~~
 
-## Client Request Behavior
+# Client Request Behavior {#request}
 
-Whenever a client wants to send QUIC packets through the proxy, or receive
-QUIC packets via the proxy, it sends a new CONNECT-QUIC request.
+A clients sends new CONNECT-QUIC requests when it wants to start
```
LPardue, line 219:
```suggestion
A client sends new CONNECT-QUIC requests when it wants to start
```

DavidSchinazi (approved):
Thanks for the discussion and edits! Approved modulo a few small suggestions

PR #31, opened 2020-11-09
Add hacky interop parser.

chris-wood:
This spits out:

```
├── interop
│   └── tests
│       ├── connect-udp.mjs
│       ├── draft-pauly-masque-quic-proxy.mjs
│       └── index.mjs
```

With contents:

```
export default
{
    "description": "TBD", 
    "tests": [
        {
            "kind": "must", 
            "name": "Client sends `Client-Connection-Id` in first `CONNECT-UDP` request", 
            "id": "connect-udp-quic-client-cid"
        }
    ], 
    "name": "QUIC-Aware Proxying Using CONNECT-UDP", 
    "id": "draft-pauly-masque-quic-proxy.mjs"
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,3 +9,9 @@ else
 	git clone -q --depth 10 $(CLONE_ARGS) \
 	    -b main https://github.com/martinthomson/i-d-template $(LIBDIR)
 endif
+
+.PHONY: interop
+interop:
+	@err=0; for f in $(drafts_source); do \
+		python interop/parse_interop_tags.py $$f interop/tests/; \
+	done
--- a/interop/parse_interop_tags.py
+++ b/interop/parse_interop_tags.py
@@ -3,24 +3,6 @@ import os
 import sys
 import json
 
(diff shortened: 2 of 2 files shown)


DavidSchinazi:
This all seems fine, but why is it in the repository for this individual draft? Shouldn't we have an official MASQUE WG repo that can hold things like this for CONNECT-UDP interop results?

chris-wood:
> This all seems fine, but why is it in the repository for this individual draft? Shouldn't we have an official MASQUE WG repo that can hold things like this for CONNECT-UDP interop results?

Maybe, though it seems simpler to manage interop requirements and results right alongside the document source. This change extracts requirements from the document source and turns them into test targets. Implementations then submit PRs indicating their conformance against those targets.

I expect the actual mechanics here to change as we get experience and feedback on how good, bad, ugly this is.

tfpauly (approved):
Looks good! Added a provisional description.

PR #60, opened 2022-08-31
Virtual Target Connection ID

ehaydenr:


draft-pauly-masque-quic-proxy.md
```diff
@@ -135,6 +133,28 @@ the client.
...
 the target.
+- Virtual Target Connection ID: a fake QUIC Connection ID that is chosen by the
+proxy that the client MUST use when sending QUIC packets in forwarding mode.
+
+## Virtual Target Connection ID
+
+QUIC allows endpoints of a connection to dictate the Connection ID their peer
+uses when sending packets. QUIC load balancing strategies such as those
+described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}} may depend on encoding
+routing information in the Connection ID. When operating in forwarding mode,
```
LPardue, line 144: I might say here something that sounds less strict

"QUIC endpoints issue Connection IDs to their peers, which are used in the Destination Connection ID field of packet sent by the peer. An endpoint on might rely on a load balancer to route QUIC packets to it, based on encoding information into the Connection IDs it issues; such as described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}. Such deployments are typically private and deployment specific and, crucially, rely on the peer using the Connection IDs that were issued. When operating in forwarding mode..."

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,6 +133,28 @@ the client.
...
+clients send QUIC packets destined for the Target directly to the Proxy. Since
+these packets are generated using the Target Connection ID, load balancers may not be
+able to route packets to the correct Proxy.
```
LPardue, line 147: How about something like

"client send QUIC packets destined for the Target, directly to the Proxy. These packets
use Connection IDs issued by the Target, which may fail to meet the exceptions of a
load balancer in front of the Proxy, introducing a potential for packet delivery failure."

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,6 +133,28 @@ the client.
...
+
+The Virtual Target Connection ID is a Proxy-dictated Connection ID the Client
+uses when sending forwarded mode packets to the Proxy. The Proxy replaces the
+Virtual Target Connection ID with the Target Connection ID prior to
+forwarding the packet to the Target. This is only necessary in the
+Client->Target direction.
```
LPardue, line 153: In keeping with prior suggestion, something like this maybe? Do we want to make this a 1:1 mapping, or allow some form of multiplicity.

"The Virtual Target Connection ID can be used to address problems with load balancer routing. A Proxy can issue a Virtual Target Connection ID to a client. When the Proxy receives a QUIC packet with a Virtual Target Connection ID value, it substitutes the value with the actual Target Connection ID prior to forwarding. Routing problems are typically only a problem for packets travelling from Client to Target, so Virtual Target Connection IDs are only used in this direction."

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,6 +133,28 @@ the client.
...
+
+Clients and Proxies not implementing forwarding mode do not need to consider
+the Virtual Target Connection ID since all Client->Target datagrams will be
+encapsulated within the Client<->Proxy connection.
```
LPardue, line 157: Can probably simplify this, when forwarding mode isn't in use, we're just back to normal UDP proxying (I think). So maybe

"When forwarding mode is not used to proxy QUIC, the concerns related to packet routing do not apply."

draft-pauly-masque-quic-proxy.md
ehaydenr: Adjusted this text to be a little more similar to [another place in the document](https://github.com/tfpauly/quic-proxy/blob/main/draft-pauly-masque-quic-proxy.md#adding-new-client-connection-ids)
> Since QUIC connection IDs are chosen by the receiver

draft-pauly-masque-quic-proxy.md
```diff
@@ -285,6 +304,54 @@ Connection ID:
...
+: A connection ID being acknowledged whose length is equal to Connection ID
+Length. This is the real Target Connection ID.
+
+Virtual Target Connection ID Length
+: The length of the connection ID being provided to the client. This must be a
+valid connection ID length for the QUIC version used in the client<->proxy QUIC
+connection. When forwarding mode is not negotiated, the length MUST be zero.
+
+Virtual Target Connection ID
+: The Proxy chosen connection ID that the client MUST use when sending packets
```
tfpauly, line 339:
```suggestion
: The Proxy-chosen connection ID that the client MUST use when sending packets
```

Issue #61, opened 2022-09-12
Consider virtual CIDs for traffic from target to client


tfpauly:
To improve CID privacy when migrating, etc, and to have more parallelism between client-to-target and target-to-client, we should look at having virtual CIDs for both directions.

tfpauly:
@ehaydenr

PR #62, opened 2022-09-19
bidirectional virtual connection ids

ehaydenr:
addresses #61

### DIFF
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -371,9 +371,9 @@ The Virtual Connection ID Length and Connection ID Length SHOULD be equal
 when possible to avoid the need to resize packets during replacement.
 
 Virtual Connection ID
-: The receiver-chosen connection ID that the sender MUST use when sending
-packets in forwarding mode. The proxy rewrites forwarding mode packets to
-contain the correct Connection ID prior to forwarding them.
+: The peer-chosen connection ID that the sender of forwarded mode packets MUST
+use when sending. The endpoint rewrites forwarding mode packets to contain the
+correct Connection ID prior to sending them.
 
 Stateless Reset Token Length
 : The length of the stateless reset token that may be sent by the client or
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -152,14 +152,14 @@ by the Proxy that the Client uses when sending forwarded mode packets. The Proxy
 replaces the Virtual Target Connection ID with the Target Connection ID prior to
 forwarding the packet to the Target.
 
-Similarly, clients supporting QUIC connection migration require that connection
-IDs aren't reused over multiple network paths to avoid linkability. The Virtual
-Client Connection ID is a connection ID chosen by the Client that the Proxy uses
-when sending forwarded mode packets. The Proxy replaces the Client Connection ID
-with the Virtual Client Connection ID prior to forwarding the packet to the
-Client. Clients take advantage of this to avoid linkability when migrating a
-client to proxy network path. Without the Virtual Client Connection ID, QUIC
-packets destined for the client would have the same connection ID.
+Similarly, QUIC requires that connection IDs aren't reused over multiple network
+paths to avoid linkability. The Virtual Client Connection ID is a connection ID
+chosen by the Client that the Proxy uses when sending forwarded mode packets.
+The Proxy replaces the Client Connection ID with the Virtual Client Connection
+ID prior to forwarding the packet to the Client. Clients take advantage of this
+to avoid linkability when migrating a client to proxy network path. The Virtual
+Client Connection ID allows the connection ID bytes to change on the wire
+without requiring the connection IDs on the client to target connection change.
 
 Clients and Proxies not implementing forwarding mode do not need to consider
 Virtual Connection IDs since all Client<->Target datagrams will be encapsulated
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -264,6 +264,17 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
 requests for very short Connection IDs as conflicts, in anticipation of future
 conflicts.
 
+## Stateless Resets for Forwarded Mode QUIC Packets
+
+While the lifecycle of forwarding rules are bound to the lifecycle of the
+client<->proxy HTTP stream, a peer may not be aware that the stream has
+terminated. If the above mappings are lost or removed without the peer's
+knowledge, they may send forwarded mode packets even though the Client
+or Proxy no longer has state for that connection. To allow the Client or
+Proxy to reset the client<->target connection in the absence the mappings
+above, a stateless reset token corresponding to the Virtual Connection ID
+may be provided.
+
 # Connection ID Capsule Types
 
 Proxy awareness of QUIC Connection IDs relies on using capsules ({{HTTP-DGRAM}})
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -341,13 +341,12 @@ Connection ID
 Length. This is the real Target or Client Connection ID.
 
 Virtual Connection ID Length
-: The length of the virtual connection ID being provided. This must be a valid
+: The length of the virtual connection ID being provided. This MUST be a valid
 connection ID length for the QUIC version used in the client<->proxy QUIC
 connection. When forwarding mode is not negotiated, the length MUST be zero.
-The Connection ID Length and Virtual Connection ID Length MAY differ in
-ACK_TARGET_CID capsules, but MUST be equal in REGISTER_CLIENT_CID capsules.
-This length equality constraint is more easily satisfied by the client since it
-controls both the Connection ID and Virtual Connection ID.
+The Virtual Connection ID Length MUST be at least as large as the Connection ID
+Length to prevent exposing the Connection ID and the lengths SHOULD be equal
+when possible to avoid the need to resize packets during replacement.
 
 Virtual Connection ID
 : The receiver-chosen connection ID that the sender MUST use when sending
@@ -469,34 +468,17 @@ When forwarding, the client sends a QUIC packet with the Virtual Target
 Connection ID in the QUIC short header, using the same socket between client and
 proxy that was used for the main QUIC connection between client and proxy.
 
-If the Virtual Target Connection ID is smaller than the Target Connection ID,
-the client MUST only write the Virtual Target Connection ID bytes over the start
-of the Target Connection ID, leaving the remainder of the Target Connection ID
-unmodified.
-
-~~~
-| 0  | 1  | 2  | 3  | 4  | 5  | 6  | 7  |
-
-  aa   bb   cc   aa   bb   cc   aa  bb    <--- Target Connection ID
-  11   22   33                            <--- Virtual Target Connection ID
-  11   22   33   aa   bb   cc   aa  bb    <--- Resulting Connection ID Bytes
-~~~
-
-If the Virtual Target Connection ID is larger than the Target Connection ID,
-the client MUST extend the length of the packet by the difference between the
-two lengths, to include the entire Virtual Target Connection ID.
-
-Clients supporting forwarding mode MUST be able to handle Virtual Target
-Connection IDs of different lengths than the corresponding Target Connection
-IDs.
-
 When forwarding, the proxy sends a QUIC packet with the Virtual Client Target
 Connection ID in the QUIC short header, using the same socket between client
 and proxy that was used for the main QUIC connection between client and proxy.
 
-The Virtual Client Connection ID is the exact same length as the Client
-Connection ID. The Proxy MUST replace the entire Virtual Client Connection ID
-with the Client Connection ID prior to sending.
+Prior to sending a forwarded mode packet, the send must replace the Connection
+ID with the Virtual Connection ID. If the Virtual Connection ID is larger than
+the Connection ID, the sender MUST extend the length of the packet by the
+difference between the two lengths, to include the entire Virtual Connection ID.
+
+Clients and proxies supporting forwarding mode MUST be able to handle Virtual
+Connection IDs of larger lengths than the corresponding Connection IDs.
 
 ## Receiving With Forwarded Mode
 
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -159,7 +159,7 @@ when sending forwarded mode packets. The Proxy replaces the Client Connection ID
 with the Virtual Client Connection ID prior to forwarding the packet to the
 Client. Clients take advantage of this to avoid linkability when migrating the
 client to proxy network path. Without the Virtual Client Connection ID, QUIC
-packets destinted for the client would have the same connection ID.
+packets destined for the client would have the same connection ID.
 
 Clients and Proxies not implementing forwarding mode do not need to consider
 Virtual Connection IDs since all Client<->Target datagrams will be encapsulated
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
 the target.
 - Virtual Target Connection ID: a fake QUIC Connection ID that is chosen by the
 proxy that the client MUST use when sending QUIC packets in forwarding mode.
+- Virtual Client Connection ID: a fake QUIC Connection ID that is chosen by the
+client that the proxy MUST use when sending QUIC packets in forwarding mode.
 
-## Virtual Target Connection ID
+## Virtual Connection IDs
 
-QUIC allows each endpoint of a connection to choose the connection IDs they
-receive with. Load balancing strategies such as those described in
-{{?QUIC-LB=I-D.ietf-quic-load-balancers}} may choose to take advantage of this
-by encoding routing information in the connection ID. When operating in
-forwarding mode, clients send QUIC packets destined for the Target directly to
-the Proxy. Since these packets are generated using the Target Connection ID,
-load balancers may not be able to route packets to the correct Proxy.
+QUIC allows each endpoint of a connection to choose the connection IDs it
+receives with. Servers receiving QUIC packets may employ load balancing
+strategies such as those described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}
+which may choose to take advantage of this by encoding routing information in
+the connection ID. When operating in forwarding mode, clients send QUIC packets
+destined for the Target directly to the Proxy. Since these packets are generated
+using the Target Connection ID, load balancers may not be able to route packets
+to the correct Proxy. The Virtual Target Connection ID is a connection ID chosen
+by the Proxy that the Client uses when sending forwarded mode packets. The Proxy
+replaces the Virtual Target Connection ID with the Target Connection ID prior to
+forwarding the packet to the Target.
 
-The Virtual Target Connection ID is a connection ID chosen by the Proxy that the
-Client uses when sending forwarded mode packets. The Proxy replaces the Virtual
-Target Connection ID with the Target Connection ID prior to forwarding the
-packet to the Target. This is only necessary in the Client->Target direction
-because the Proxy is otherwise the only receiver of QUIC packets with connection
-IDs it did not generate.
+Similarly, clients supporting QUIC connection migration require that connection
+IDs aren't reused over multiple network paths to avoid linkability. The Virtual
+Client Connection ID is a connection ID chosen by the Client that the Proxy uses
+when sending forwarded mode packets. The Proxy replaces the Client Connection ID
+with the Virtual Client Connection ID prior to forwarding the packet to the
+Client. Clients take advantage of this to avoid linkability when migrating the
+client to proxy network path. Without the Virtual Client Connection ID, QUIC
+packets destinted for the client would have the same connection ID.
 
 Clients and Proxies not implementing forwarding mode do not need to consider
-the Virtual Target Connection ID since all Client->Target datagrams will be
-encapsulated within the Client<->Proxy connection.
+Virtual Connection IDs since all Client<->Target datagrams will be encapsulated
+within the Client<->Proxy connection.
 
 # Required Proxy State {#mappings}
 
@@ -213,11 +221,12 @@ does not allow forwarded mode does not need to maintain this mapping.
 Each pair of Client Connection ID and target-facing socket MUST map to a single
 stream on a single client <-> proxy HTTP stream. Additionally, when supporting
 forwarding mode, the pair of Client Connection ID and target-facing socket MUST
-map to a single client-facing socket.
+map to a single client-facing socket and Virtual Client Connection ID.
 
 ~~~
 (Target-facing socket + Client Connection ID) => (Client <-> Proxy HTTP Stream)
-(Target-facing socket + Client Connection ID) => Client-facing socket
+(Target-facing socket + Client Connection ID)
+    => (Client-facing socket + Virtual Client Connection ID)
 ~~~
 
 Multiple pairs of Connection IDs and target-facing sockets can map to the same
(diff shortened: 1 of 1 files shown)


draft-pauly-masque-quic-proxy.md
```diff
@@ -328,29 +338,33 @@ bytes, but QUIC invariants allow up to 255 bytes.
...
 
-Virtual Target Connection ID Length
-: The length of the connection ID being provided to the client. This must be a
-valid connection ID length for the QUIC version used in the client<->proxy QUIC
+Virtual Connection ID Length
+: The length of the virtual connection ID being provided. This must be a valid
+connection ID length for the QUIC version used in the client<->proxy QUIC
 connection. When forwarding mode is not negotiated, the length MUST be zero.
+The Connection ID Length and Virtual Connection ID Length MAY differ in
+ACK_TARGET_CID capsules, but MUST be equal in REGISTER_CLIENT_CID capsules.
```
DavidSchinazi, line 348: Why the discrepancy here? Is there a reason to force the lengths to be equal?

draft-pauly-masque-quic-proxy.md
```diff
@@ -328,29 +338,33 @@ bytes, but QUIC invariants allow up to 255 bytes.
...
+This length equality constraint is more easily satisfied by the client since it
+controls both the Connection ID and Virtual Connection ID.
 
-Virtual Target Connection ID
-: The Proxy-chosen connection ID that the client MUST use when sending packets
-in forwarding mode. The proxy rewrites forwarding mode packets to contain the
-correct Target Connection ID prior to forwarding them on to the Target.
+Virtual Connection ID
+: The recevier-chosen connection ID that the sender MUST use when sending
```
DavidSchinazi, line 353:
```suggestion
: The receiver-chosen connection ID that the sender MUST use when sending
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -328,29 +338,33 @@ bytes, but QUIC invariants allow up to 255 bytes.
...
+The Connection ID Length and Virtual Connection ID Length MAY differ in
+ACK_TARGET_CID capsules, but MUST be equal in REGISTER_CLIENT_CID capsules.
+This length equality constraint is more easily satisfied by the client since it
+controls both the Connection ID and Virtual Connection ID.
 
-Virtual Target Connection ID
-: The Proxy-chosen connection ID that the client MUST use when sending packets
-in forwarding mode. The proxy rewrites forwarding mode packets to contain the
-correct Target Connection ID prior to forwarding them on to the Target.
+Virtual Connection ID
```
DavidSchinazi, line 352: Mentions of sender and receiver in this paragraph are unclear: I was expecting them to be the sender/receiver of the capsule being discussed, but that doesn't appear to be the case

draft-pauly-masque-quic-proxy.md
```diff
@@ -328,29 +338,33 @@ bytes, but QUIC invariants allow up to 255 bytes.
...
+packets in forwarding mode. The proxy rewrites forwarding mode packets to
+contain the correct Connection ID prior to forwarding them.
 
 Stateless Reset Token Length
-: The length of the stateless reset token that may be sent by the proxy
-in response to forwarded mode packets in order to reset the Client<->Target
-QUIC connection. When forwarding mode is not negotiated, the length MUST be
-zero. Proxies choosing not to support stateless resets MAY set the length to
-zero. Clients receiving a zero-length stateless reset token MUST ignore it.
+: The length of the stateless reset token that may be sent by the client or
```
DavidSchinazi, line 358: This isn't specific to this PR, but it would be nice to add a paragraph explaining why proxies care about stateless reset tokens and in which scenarios this feature would be useful

DavidSchinazi (commented):
Thanks for doing this! A few minor comments

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
-the Proxy. Since these packets are generated using the Target Connection ID,
-load balancers may not be able to route packets to the correct Proxy.
-
-The Virtual Target Connection ID is a connection ID chosen by the Proxy that the
-Client uses when sending forwarded mode packets. The Proxy replaces the Virtual
-Target Connection ID with the Target Connection ID prior to forwarding the
-packet to the Target. This is only necessary in the Client->Target direction
-because the Proxy is otherwise the only receiver of QUIC packets with connection
-IDs it did not generate.
+- Virtual Client Connection ID: a fake QUIC Connection ID that is chosen by the
```
tfpauly, line 138: To match the order of the CIDs above, let's put Virtual CCID before Virtual TCID

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
+client that the proxy MUST use when sending QUIC packets in forwarding mode.
+
+## Virtual Connection IDs
+
+QUIC allows each endpoint of a connection to choose the connection IDs it
+receives with. Servers receiving QUIC packets may employ load balancing
```
ekinnear, line 144:
```suggestion
receives. Servers receiving QUIC packets may employ load balancing
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
-Target Connection ID with the Target Connection ID prior to forwarding the
-packet to the Target. This is only necessary in the Client->Target direction
-because the Proxy is otherwise the only receiver of QUIC packets with connection
-IDs it did not generate.
+- Virtual Client Connection ID: a fake QUIC Connection ID that is chosen by the
+client that the proxy MUST use when sending QUIC packets in forwarding mode.
+
+## Virtual Connection IDs
+
+QUIC allows each endpoint of a connection to choose the connection IDs it
```
ekinnear, line 143:
```suggestion
QUIC allows each endpoint of a connection to choose the destination connection IDs it
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
+the connection ID. When operating in forwarding mode, clients send QUIC packets
+destined for the Target directly to the Proxy. Since these packets are generated
+using the Target Connection ID, load balancers may not be able to route packets
+to the correct Proxy. The Virtual Target Connection ID is a connection ID chosen
+by the Proxy that the Client uses when sending forwarded mode packets. The Proxy
+replaces the Virtual Target Connection ID with the Target Connection ID prior to
+forwarding the packet to the Target.
+
+Similarly, clients supporting QUIC connection migration require that connection
+IDs aren't reused over multiple network paths to avoid linkability. The Virtual
```
ekinnear, line 156: I think this is true for all clients, can we just say that "QUIC requires that connections IDs aren't reused over multiple network paths to..."?

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
+Client Connection ID is a connection ID chosen by the Client that the Proxy uses
+when sending forwarded mode packets. The Proxy replaces the Client Connection ID
+with the Virtual Client Connection ID prior to forwarding the packet to the
+Client. Clients take advantage of this to avoid linkability when migrating the
+client to proxy network path. Without the Virtual Client Connection ID, QUIC
```
ekinnear, line 161: To "a" proxy network path? To "the" proxy network path?

draft-pauly-masque-quic-proxy.md
```diff
@@ -135,27 +135,35 @@ is used in the Destination Connection ID field of packets from the client to
...
+packets destinted for the client would have the same connection ID.
```
ekinnear, line 162:
```suggestion
packets destined for the client would have the same connection ID.
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -474,6 +490,14 @@ Clients supporting forwarding mode MUST be able to handle Virtual Target
 Connection IDs of different lengths than the corresponding Target Connection
 IDs.
 
+When forwarding, the proxy sends a QUIC packet with the Virtual Client Target
+Connection ID in the QUIC short header, using the same socket between client
```
ekinnear, line 494: Same socket? Same 5-tuple?

draft-pauly-masque-quic-proxy.md
ehaydenr: I was thinking about differences between passive and active migration. In the passive case, the CID may not have changed despite having been reused over a new network path. I'm wondering if it is necessary to mention this here or if the language you propose is sufficient.

ekinnear: QUIC has text about that already, I'd be tempted to say that we just refer to that since it was very carefully worded 🤷

draft-pauly-masque-quic-proxy.md
ehaydenr: The discrepancy is motivated by implementation considerations. If implementing forwarding in a way that requires fixing IP and/or UDP checksums, only length changes by a multiple of 4 can be done incrementally. If the length difference is not a multiple of 4, the entire checksum would have to be recomputed. One example of this constraint is documented in [Linux's BPF helpers](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) (bpf_csum_diff):

> Compute a checksum difference, from the raw buffer
                     pointed by from, of length from_size (that must be
                     a multiple of 4), towards the raw buffer pointed by
                     to, of size to_size (same remark).

In the Virtual Target Connection ID case, the requirement to support Virtual Connection IDs of a different length is quite clear. Load balancing strategies such as those described QUIC-LB may require a minimum CID length to encode information (e.g. 17 bytes in the [single-pass encryption](https://datatracker.ietf.org/doc/html/draft-ietf-quic-load-balancers#section-4.3.1) case). If the proxy intended on using an incremental checksum algorithm, it could make sure that the Virtual Target CID is a multiple of 4 larger than the Target CID. For example, if the Target CID is 15 bytes, it could provide a Virtual Target CID of 19 bytes (17 "real" bytes encoding/encrypting information followed by 2 randomly chosen bytes). The replace operation of a 19 bytes buffer to a 15 byte buffer would be compatible with incremental checksum algorithms since 4 bytes are removed.

Requiring the Target Connection ID be at least the length of the Virtual Target Connection ID would necessitate cooperation with the target and as such is a non-starter. The draft currently allows any type of length difference between the Target Connection ID and Virtual Target Connection ID. We could instead tighten this up by requiring that the Virtual Target Connection ID be at least as long as the Target Connection ID, but I'm not sure how useful this is.

By requiring that the Client CID and Virtual Client CID lengths be equal, we're guaranteeing that incremental checksum is possible. We could instead require that the difference in length between Client CID and Virtual Client CID be a multiple of 4, but that seems like too specific of a requirement - particularly if it is of no real consequence to simply require they be equal.

By taking the above Virtual Target Connection ID length difference case as a requirement, we're precluding the client from using incremental checksum algorithms in the first place because it must be able to translate between arbitrary lengths when sending. Given that the client already needs to handle arbitrary length differences and that Client and Virtual Client CIDs are BOTH generated by the client, I would think that it has sufficient flexibility in crafting a Virtual Client CID of equivalent length that meet its needs. If we chose not to require this, the inclusion of Virtual Client CID (this PR) would introduce the requirement to do a full checksum recalculation for the reasons described above.

draft-pauly-masque-quic-proxy.md
ehaydenr: I'm reusing the language as above: https://github.com/tfpauly/quic-proxy/blob/050afbdca7d33c327b67a56d6a9ffe5a0353b988/draft-pauly-masque-quic-proxy.md?plain=1#L453 `git blame` shows @DavidSchinazi so I'll defer to him

draft-pauly-masque-quic-proxy.md
ehaydenr: > ... choose the destination connection IDs it receives

To me, that reads a little ambiguously. An endpoints of a QUIC connection "receives" connection IDs from its peer via NEW_CONNECTION_ID frames, but that's obviously not what you meant. That being said, I don't love the language as-is either, but it's a carry over from the previous PR. May or may not make sense to address it now.

draft-pauly-masque-quic-proxy.md
ehaydenr: good catch: 5c419f6be0ae9b855be73a397dc114a6085bb829

draft-pauly-masque-quic-proxy.md
ehaydenr: thanks, updated the language and made a few other changes here: 65db5de9489c96463f48ab2d9d8edafdfc8f1d8f

draft-pauly-masque-quic-proxy.md
DavidSchinazi: Thanks for the detailed explanation. I would suggest making this a SHOULD instead of a MUST: if the absolute worst case scenario is recomputing the checksum on each packet, then that doesn't warrant a MUST. There might be use-cases where a client or proxy isn't in control of its CID length, and in such scenarios they'd rather pay the checksum cost instead of failing.

draft-pauly-masque-quic-proxy.md
DavidSchinazi: Sockets are explicitly defined as UDP 4-tuples in Section 1.2

draft-pauly-masque-quic-proxy.md
ehaydenr: Something I forgot to include that is also worth considering is that the Virtual Client CID must be at least as long as the Client CID to avoid linkability. Consider the case of connection migration. If the Virtual Client CID is 0xaabb and the Client CID is 0x11223344, and the Virtual Client CID was changed due to migration to 0xccdd, the packet would go from flags byte followed by 0xaabb3344 to 0xccdd3344. The "3344" would increase linkability.

The Virtual Target Connection ID has the same issue - the client<->proxy network path change would expose any extra client<->target CID bytes. I think in both cases the Virtual Connection ID must be at least as long as its counterpart.

ehaydenr: Addressed here: 6cad095ceea2b5d54233a5111253a2351a894149

draft-pauly-masque-quic-proxy.md
ehaydenr: Agreed, how about something like this?

> The peer-chosen connection ID that the sender of forwarded mode packets MUST
> use when sending. The endpoint rewrites forwarding mode packets to contain the
> correct Connection ID prior to sending them.

draft-pauly-masque-quic-proxy.md
ehaydenr: Great point, I added some language here: f6ab95a1c0313babf5eb4c29e8fad8485fb0cab6

draft-pauly-masque-quic-proxy.md
DavidSchinazi: Oh this doesn't match my understanding of how replacement works, which makes me wonder if the replacement is sufficiently well documented. I was assuming that in this scenario you would replace `0x11223344` with `0xaabb` (not `0xaabb3344`) so you might see these two packets in sequence:
* before replacement `0x11223344abcdef...` `0x11223344fedcba...`
* after replacement `0xaabbabcdef...` `0xaabbfedcba...`

Am I misunderstanding how replacement works?

draft-pauly-masque-quic-proxy.md
```diff
@@ -255,6 +264,17 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
...
+## Stateless Resets for Forwarded Mode QUIC Packets
+
+While the lifecycle of forwarding rules are bound to the lifecycle of the
+client<->proxy HTTP stream, a peer may not be aware that the stream has
+terminated. If the above mappings are lost or removed without the peer's
+knowledge, they may send forwarded mode packets even though the Client
+or Proxy no longer has state for that connection. To allow the Client or
+Proxy to reset the client<->target connection in the absence the mappings
+above, a stateless reset token corresponding to the Virtual Connection ID
+may be provided.
```
DavidSchinazi, line 276: This new text is great, can we also augment it with an example where one device reboots and loses state, and what happens in that scenario to properly close the connection?

draft-pauly-masque-quic-proxy.md
ehaydenr: In a previous commit, the replacement where the virtual cid is smaller than the cid is described [here](https://github.com/tfpauly/quic-proxy/pull/62/commits/6cad095ceea2b5d54233a5111253a2351a894149#diff-4dd090ab6ff7820385219d3ec1e86133930ab3a3b71aee2fe6e1a61a459f8da4L472)

> If the Virtual Target Connection ID is smaller than the Target Connection ID,
 the client MUST only write the Virtual Target Connection ID bytes over the start
 of the Target Connection ID, leaving the remainder of the Target Connection ID
 unmodified.

That was removed in 6cad095ceea2b5d54233a5111253a2351a894149 when changing the requirement to be that the virtual cid be at least as long as the cid.

I can update to document each case (virtual smaller, equal, and larger) to try to make this clearer once we get some alignment on what (if any) requirements should exist and what replacement should look like.

I hadn't considered shrinking the packet when the virtual cid is smaller than the cid, but it would satisfy the exposure issue during migration and should work equally well. Assuming we go with this approach, the replacement is always a "full" replacement. @tfpauly not sure if you have any opinions here.

draft-pauly-masque-quic-proxy.md
ehaydenr: Thanks, I added an example in e191d7d958c2f127038f8df6426917c6958a763f. I went with a slightly different example, but I think it communicates the same thing.

DavidSchinazi: Great, thanks!

draft-pauly-masque-quic-proxy.md
ehaydenr: @DavidSchinazi removed length restrictions in 02f2fd47be22fcfba66c5d085d9406bf3e752f4c

draft-pauly-masque-quic-proxy.md
ehaydenr: updated with the above language in 7a2ec8dfe7b644a8f93e903a2ef500d6d49972c5

draft-pauly-masque-quic-proxy.md
tfpauly: Yeah, let's not address for now.

draft-pauly-masque-quic-proxy.md
```diff
@@ -133,29 +133,37 @@ the client.
...
+## Virtual Connection IDs
+
+QUIC allows each endpoint of a connection to choose the connection IDs it
+receives with. Servers receiving QUIC packets may employ load balancing
+strategies such as those described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}
+which may choose to take advantage of this by encoding routing information in
+the connection ID. When operating in forwarding mode, clients send QUIC packets
+destined for the Target directly to the Proxy. Since these packets are generated
+using the Target Connection ID, load balancers may not be able to route packets
+to the correct Proxy. The Virtual Target Connection ID is a connection ID chosen
```
tfpauly, line 150:
```suggestion
using the Target Connection ID, load balancers would not be able to route packets
to the correct Proxy if the packets were sent with the Target Connection ID.
The Virtual Target Connection ID is a connection ID chosen
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -133,29 +133,37 @@ the client.
...
-The Virtual Target Connection ID is a connection ID chosen by the Proxy that the
-Client uses when sending forwarded mode packets. The Proxy replaces the Virtual
-Target Connection ID with the Target Connection ID prior to forwarding the
-packet to the Target. This is only necessary in the Client->Target direction
-because the Proxy is otherwise the only receiver of QUIC packets with connection
-IDs it did not generate.
+## Virtual Connection IDs
+
+QUIC allows each endpoint of a connection to choose the connection IDs it
+receives with. Servers receiving QUIC packets may employ load balancing
```
tfpauly, line 144:
```suggestion
receives with. Servers receiving QUIC packets can employ load balancing
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -133,29 +133,37 @@ the client.
...
+strategies such as those described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}
+which may choose to take advantage of this by encoding routing information in
```
tfpauly, line 146:
```suggestion
that encode routing information in
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -482,10 +512,11 @@ it has a valid mapping.
 
 Once a client has sent "Proxy-QUIC-Forwarding" with a value of "?1", it MUST be
 prepared to receive forwarded short header packets on the socket between itself
-and the proxy for any Client Connection ID that it has registered with a
+and the proxy for any Virtual Client Connection ID that it has registered with a
 REGISTER_CLIENT_CID capsule. The client uses the Destination Connection ID field
 of the received packet to determine if the packet was originated by the proxy,
-or merely forwarded from the target.
+or merely forwarded from the target. The client replaces the Virtual Client
+Connection ID with the real Client Connection ID.
```
tfpauly, line 519:
```suggestion
Connection ID with the real Client Connection ID before processing the packet further.
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -304,22 +337,22 @@ Connection ID:
 length of the capsule. Note that in QUICv1, the length of the Connection ID
 is limited to 20 bytes, but QUIC invariants allow up to 255 bytes.
 
-The ACK_TARGET_CID capsule type includes the Virtual Target Connection ID
-and a Stateless Reset Token.
+The REGISTER_CLIENT_CID and ACK_TARGET_CID capsule types includes a Virtual
```
tfpauly, line 340:
```suggestion
The REGISTER_CLIENT_CID and ACK_TARGET_CID capsule types include a Virtual
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -255,6 +264,29 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
...
+## Stateless Resets for Forwarded Mode QUIC Packets
+
+While the lifecycle of forwarding rules are bound to the lifecycle of the
+client<->proxy HTTP stream, a peer may not be aware that the stream has
+terminated. If the above mappings are lost or removed without the peer's
+knowledge, they may send forwarded mode packets even though the Client
+or Proxy no longer has state for that connection. To allow the Client or
+Proxy to reset the client<->target connection in the absence the mappings
+above, a stateless reset token corresponding to the Virtual Connection ID
+may be provided.
```
tfpauly, line 276:
```suggestion
can be provided.
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -255,6 +264,29 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
...
+
+Consider a proxy that initiates closure of a client<->proxy QUIC connection.
+If the client is temporarily unresponsive or unreachable, the proxy may have
```
tfpauly, line 279:
```suggestion
If the client is temporarily unresponsive or unreachable, the proxy might have
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -255,6 +264,29 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
...
+considered the connection closed and removed all connection state (including
+the above stream mappings). If the client never learned about the closure, it
```
tfpauly, line 281:
```suggestion
the stream mappings used for forwarding). If the client never learned about the closure, it
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -255,6 +264,29 @@ particularly zero-length Connection IDs, a proxy MAY choose to reject all
...
+might send forwarded mode packets to the proxy, assuming the stream mappings
+and client<->proxy connection are still intact. The proxy receives these
+forwarding mode packets, but doesn't have any state corresponding to the
```
tfpauly, line 284:
```suggestion
and client<->proxy connection are still intact. The proxy will receive these
forwarding mode packets, but won't have any state corresponding to the
```

Issue #48, opened 2021-07-12
Stateless Reset handling


MikeBishop:
The draft doesn't currently discuss Stateless Resets at all.

When using vanilla CONNECT-UDP, each client has a dedicated origin-facing port.  If the origin sends a Stateless Reset, it will be forwarded naively, parsed by the client, and the client will close the stream as it is no longer needed.  (Or the client could start a new connection using the same tunnel; doesn't matter to the proxy.)

When using forwarding mode, the proxy will reject Stateless Reset packets as they do not match any open request's registered CIDs.  The client doesn't learn that the connection has failed and will continue to send packets, which the origin will continue to reject, until the connection eventually times out.

Should we consider having the client provide Stateless Reset tokens to the server alongside the CIDs?  In that case, the proxy could either correctly relay Stateless Reset packets or convert them to stream error(s) for the affected CID.

PR #66, opened 2022-10-30
client communication of target-provided stateless reset token

ehaydenr:
Addresses #48, however, instead of converting stateless reset packets to a stream error, they are simply tunneled or forwarded to the client.

A couple questions come to mind:

**Should receipt of a stateless reset packet result in a stream error?**

- Offers reliably delivered signal to the client that the connection is broken.
- Allows the proxy to terminate the stream immediately and avoid having to consider handling of many stateless reset packets
- What H3 error code? `H3_CONNECT_ERROR`?

**If not a stream error, is tunneled _OR_ forwarded OK?**

- If we require they be tunneled, does that make things easier for the client at no cost to the proxy?
- Is there any reason why we wouldn't want the stateless reset packet sent "in the clear" (forwarded)?

### DIFF
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -363,7 +363,7 @@ Register Target Connection ID Capsule {
   Stateless Reset Token (..),
 }
 ~~~
-{: #fig-capsule-cid title="Register Target Connection ID Capsule Format"}
+{: #fig-capsule-register-target-cid title="Register Target Connection ID Capsule Format"}
 
 Connection ID Length
 : The length of the connection ID being registered, which is between 0 and
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -288,6 +288,18 @@ reset token for the Virtual Target Connection ID, it can send a stateless reset
 packet to quickly notify the client that the client<->target connection is
 broken.
 
+## Stateless Resets from the Target
+
+Reuse of target-facing sockets is only possible because QUIC connection IDs
+allow distinguishing packets for multiple QUIC connections received with the
+same 5-tuple. One exception to this is Stateless Reset packets, in which the
+connection ID is not used, but rather populated with unpredictable bits followed
+by a Stateless Reset token, to make it indistinguishable from a regular packet
+with a short header. In order for the proxy to correctly recognize Stateless
+Reset packets, the client SHOULD share the Stateless Reset token for each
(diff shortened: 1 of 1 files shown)


tfpauly:
I lean towards forwarding the stateless reset, rather than making this a stream error. That’s effectively what the PR does, right?

ehaydenr:
> That’s effectively what the PR does, right?

Yes, that was the intention. It also allows for _either_ tunneled or forwarded. That seemed like the least prescriptive, but I can also see an argument for restricting it only being sent via the tunnel instead of being forwarded.

tfpauly:
Yeah I could see that it might interfere with forwarding. Saying that it only tunnels would be a bit simpler. Only short header packets get forwarded, and this arguably doesn't meet that bar.

draft-pauly-masque-quic-proxy.md
```diff
@@ -288,6 +288,19 @@ reset token for the Virtual Target Connection ID, it can send a stateless reset
...
+
+Reuse of target-facing sockets is only possible because QUIC connection IDs
+allow distinguishing packets for multiple QUIC connections received with the
+same 5-tuple. One exception to this is Stateless Reset packets, in which the
+connection ID is not used, but rather populated with unpredictable bits followed
+by a Stateless Reset token, to make it indistinguishable from a regular packet
+with a short header. In order for the proxy to correctly recognize Stateless
+Reset packets, the client SHOULD share the Stateless Reset token for each
+registered Target Connection ID. When the proxy receives a Stateless Reset packet,
+it SHOULD tunnel the packet to the client regardless of whether or not forwarding
```
ehaydenr, line 301: @tfpauly, is the wording here strict enough to prohibit forwarding (as opposed to tunneling) of stateless reset packets? If not, any suggested language?

draft-pauly-masque-quic-proxy.md
```diff
@@ -288,6 +288,19 @@ reset token for the Virtual Target Connection ID, it can send a stateless reset
...
+is enabled.
+
```
tfpauly, line 303:
```suggestion
it can send the packet to the client as a tunnelled datagram. Although Stateless Reset packets
look like short header packets, they are not technically short header packets and do not contain
negotiated connection IDs, and thus are not eligible for forwarded mode.
```

tfpauly: @ehaydenr how about this?

tfpauly (approved):

Issue #68, opened 2022-10-31
CID Registration doesn't consider uncooperative servers


martinduke:
I'm really happy with the virtual CID concept, and I think it will work well with the happy path. IIUC there is a 1-to-1 mapping of connection IDs to virtual connection IDs and vice versa. Thus the proxy has a very simple job of substituting one for the other.

However, there are some target behaviors where migration won't work all that well:

1) The target has disabled migration but the proxy allows it. When the client migrates, it wants to use a new (virtual) target connection ID but the new target connection ID might not be routed correctly.

2) The target is not issuing new connection IDs. Again, the client would like to use a new target virtual CID but obviously can't use a new target CID.

3) The client changes the connection ID in use, but the target doesn't observe migration (proxy IP address is the same) so it declines to use a new client connection ID (I believe this is allowed in RFC 9000). Thus proxy->client packets will use the old virtual client CID and remain linkable.

Having thought about it for a moment, the answer is probably to decouple the "actual" CIDs and the virtual ones. I think this means lots of different capsules.

REGISTER_VIRTUAL_CLIENT_CID: adds to a pool of CIDs that the proxy can choose from. When the client changes address, the proxy MUST choose a new CID from this pool (this requirement is already sorta there because the proxy can't use the same CID on two paths).

REGISTER_CLIENT_CID: As the target will not experience a migration, it's not clear to me why a client would ever send a NEW_CONNECTION_ID frame. But if it chooses to, the proxy has to know that incoming packets with that CID correspond to this stream.

[Perhaps "retire" versions of this capsule are also necessary?]

REQUEST_VIRTUAL_TARGET_CID: Regardless of whether or not the actual target CID is changing, the client needs new proxy CIDs in case the client->proxy connection migrates. This must be acked with a routable CID.

[again, maybe there has to be a proxy-sent capsule that retires these? A long lived connection might need to discard old QUIC-LB CIDs if the keys have rotated, etc]

UPDATE_TARGET_CID: I believe the proxy only needs one target CID at a time, which it applies to all outgoing packets in forwarding mode. The only reason the client should start using a new target CID is if the old one is being retired. This is not a migration and doesn't need probes, etc. 

(this Issue does not consider Multipath QUIC, which has its own set of concerns.)

martinduke:
[...]

PR #72, opened 2022-11-04
update forwarded mode migration guidance

ehaydenr:
addresses #68 by clarifying active vs. passive migration guidance and describing how to change virtual CIDs without changing real CIDs

### DIFF
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -462,14 +462,14 @@ Connection ID capsules.
 The client sends a REGISTER_CLIENT_CID capsule whenever it advertises a new
 Client Connection ID to the target, and a REGISTER_TARGET_CID capsule when
 it has received a new Target Connection ID for the target. In order to change
-the connection ID bytes on the wire, a client may change the Virtual Client
-Connection ID by sending a REGISTER_CLIENT_CID with the same Connection ID,
-but different Virtual Connection ID. Similarly, the client may solicit a new
+the connection ID bytes on the wire, a client can update a previously advertised Virtual Client
+Connection ID by sending a new REGISTER_CLIENT_CID with the same Connection ID,
+but a different Virtual Connection ID. Similarly, the client may solicit a new
 Virtual Target Connection ID by sending a REGISTER_TARGET_CID capsule with
-the previously registered Target Connection ID. Clients are responsible for
+a previously registered Target Connection ID. Clients are responsible for
 changing Virtual Connection IDs when the HTTP stream's network path changes to
-avoid linkability across network paths. Note that the initial
-REGISTER_CLIENT_CID capsule MAY be sent prior to receiving an HTTP response
+avoid linkability across network paths. Note that initial
+REGISTER_CLIENT_CID capsules MAY be sent prior to receiving an HTTP response
 from the proxy.
 
 ## New Proxied Connection Setup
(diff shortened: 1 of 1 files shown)


draft-pauly-masque-quic-proxy.md
```diff
@@ -461,9 +461,16 @@ Connection ID capsules.
...
+it has received a new Target Connection ID for the target. In order to change
+the connection ID bytes on the wire, a client may change the Virtual Client
+Connection ID by sending a REGISTER_CLIENT_CID with the same Connection ID,
+but different Virtual Connection ID. Similarly, the client may solicit a new
+Virtual Target Connection ID by sending a REGISTER_TARGET_CID capsule with
+the previously registered Target Connection ID. Clients are responsible for
+changing Virtual Connection IDs when the HTTP stream's network path changes to
+avoid linkability across network paths. Note that the initial
+REGISTER_CLIENT_CID capsule MAY be sent prior to receiving an HTTP response
+from the proxy.
```
tfpauly, line 473:
```suggestion
it has received a new Target Connection ID for the target. In order to change
the connection ID bytes on the wire, a client can update a previously advertised Virtual Client
Connection ID by sending a new REGISTER_CLIENT_CID with the same Connection ID,
but a different Virtual Connection ID. Similarly, the client may solicit a new
Virtual Target Connection ID by sending a REGISTER_TARGET_CID capsule with
a previously registered Target Connection ID. Clients are responsible for
changing Virtual Connection IDs when the HTTP stream's network path changes to
avoid linkability across network paths. Note that initial
REGISTER_CLIENT_CID capsules MAY be sent prior to receiving an HTTP response
from the proxy.
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -675,18 +682,20 @@ to an unvalidated client address to the size of an initial congestion window.
...
+an active migration, the proxy removes forwarding rules in order to not send
+packets with the same connection ID bytes over multiple network paths. After
+initiating an active migration, clients are no longer able to send forwarded mode
+packets since the proxy will have removed the rules. Clients may proceed with
+tunneled mode or may request new forwarding rules via REGISTER_CLIENT_CID and
+REGISTER_TARGET_CID capsules. Each of these capsules shall contain new virtual
+connection IDs to prevent packets with the same connection ID bytes being used
+over multiple network paths. Note that the Client Connection ID and Target
+Connection ID may stay the same while the Virtual Target Connection ID and
+Virtual Client Connection ID change.
```
tfpauly, line 698:
```suggestion
When operating in forwarded mode, the proxy reconfigures or removes forwarding
rules as the network path between the client and proxy changes. In the event of
passive migration, the proxy automatically reconfigures forwarding rules to use
the latest active and validated network path for the HTTP stream. In the event of
active migration, the proxy removes forwarding rules in order to not send
packets with the same connection ID bytes over multiple network paths. After
initiating active migration, clients are no longer able to send forwarded mode
packets since the proxy will have removed forwarding rules. Clients can proceed with
tunnelled mode or can request new forwarding rules via REGISTER_CLIENT_CID and
REGISTER_TARGET_CID capsules. Each of these capsules will contain new virtual
connection IDs to prevent packets with the same connection ID bytes being used
over multiple network paths. Note that the Client Connection ID and Target
Connection ID can stay the same while the Virtual Target Connection ID and
Virtual Client Connection ID change.
```

tfpauly (commented):
I like re-sending the REGISTER_CLIENT_CID with a new VCID, but it seems a bit odd to send a REGISTER_TARGET_CID that's duplicated as a signal to get a new VCID from the proxy.

I see a few options:
- Keep this proposal as-is
- Before re-registering, have the client send explicit CLOSE_ messages for the old ones. This gets you to a "restarted" state
- Client sends a new REGISTER_CLIENT, but not a REGISTER_TARGET, and the proxy sending a new ACK_TARGET with a new VCID uninitiated.

Not sure which I prefer...

ehaydenr:
I think one of the challenges with doing CLOSE_ first is if you do CLOSE_CLIENT_CID then the proxy thinks it needs to remove state for port reuse too

tfpauly (approved):

PR #75, opened 2023-03-10
Mention using PINGs to keep tunnels alive

### LINKED ISSUE #69: Are there potentially no ACKs? What then?
Forwarding mode, of course, doesn't have packet numbers for the outer connection. If there's nothing else going over the outer connection, it is therefore impractical for the proxy to send ACKs. So what are the implications for congestion control, idle timeouts, and so on?

Frequent ACKs in response to QUIC packets are pretty fundamental to congestion control, timeouts, and perhaps other transport layer mechanisms, and we ought to think through the implications of fully relying on the inner connection for all of those functions.

tfpauly:
Closes #69

### DIFF
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -573,9 +573,9 @@ Connection ID with the real Client Connection ID before processing the packet fu
 ## Connection Maintenance in Forwarded Mode
 
 When a client and proxy are using forwarded mode, it is possible that there can be
-long periods of time in which no stream data or datagram frames is exchanged
+long periods of time in which no ack-eliciting packets (see {{Section 2 of !QUIC-RETRANSMISSION=RFC9002}}) are exchanged
 between the client and proxy. If these periods extend beyond the max_idle_timeout
-values for the QUIC connection to the proxy (see {{Section 10.1 of QUIC}}),
+values for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
 the QUIC connection might be closed by the proxy if the proxy does not use
 forwarded packets as an explicit liveness signal. To avoid this, clients SHOULD
 send keepalive packets to the proxy before the idle timeouts would be reached,
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -574,8 +574,8 @@ Connection ID with the real Client Connection ID before processing the packet fu
 
 When a client and proxy are using forwarded mode, it is possible that there can be
 long periods of time in which no ack-eliciting packets (see {{Section 2 of !QUIC-RETRANSMISSION=RFC9002}}) are exchanged
-between the client and proxy. If these periods extend beyond the max_idle_timeout
-values for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
+between the client and proxy. If these periods extend beyond the effective idle
+timeout for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
 the QUIC connection might be closed by the proxy if the proxy does not use
 forwarded packets as an explicit liveness signal. To avoid this, clients SHOULD
 send keepalive packets to the proxy before the idle timeouts would be reached,
--- a/draft-pauly-masque-quic-proxy.md
+++ b/draft-pauly-masque-quic-proxy.md
@@ -579,7 +579,8 @@ values for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
 the QUIC connection might be closed by the proxy if the proxy does not use
 forwarded packets as an explicit liveness signal. To avoid this, clients SHOULD
 send keepalive packets to the proxy before the idle timeouts would be reached,
-which can be done using PING frames.
+which can be done using a PING frame or another ack-eliciting frame as described
+in {{Section 10.1.1 of QUIC}}.
 
 # Proxy Behavior {#proxy-behavior}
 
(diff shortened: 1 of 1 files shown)


draft-pauly-masque-quic-proxy.md
```diff
@@ -570,6 +570,17 @@ of the received packet to determine if the packet was originated by the proxy,
 or merely forwarded from the target. The client replaces the Virtual Client
 Connection ID with the real Client Connection ID before processing the packet further.
 
+## Connection Maintenance in Forwarded Mode
+
+When a client and proxy are using forwarded mode, it is possible that there can be
+long periods of time in which no stream data or datagram frames is exchanged
```
DavidSchinazi, line 576:
```suggestion
long periods of time in which no ack-eliciting packets (see {{Section 2 of !QUIC-RETRANSMISSION=RFC9002}}) are exchanged
```

draft-pauly-masque-quic-proxy.md
```diff
@@ -570,6 +570,17 @@ of the received packet to determine if the packet was originated by the proxy,
...
+between the client and proxy. If these periods extend beyond the max_idle_timeout
+values for the QUIC connection to the proxy (see {{Section 10.1 of QUIC}}),
+the QUIC connection might be closed by the proxy if the proxy does not use
+forwarded packets as an explicit liveness signal. To avoid this, clients SHOULD
+send keepalive packets to the proxy before the idle timeouts would be reached,
+which can be done using PING frames.
```
LPardue, line 582: The intent is good, but maybe you just want to defer to QUIC's liveness testing text (https://datatracker.ietf.org/doc/html/rfc9000#section-10.1.1) as to how exactly to do it?

draft-pauly-masque-quic-proxy.md
```diff
@@ -570,6 +570,17 @@ of the received packet to determine if the packet was originated by the proxy,
 or merely forwarded from the target. The client replaces the Virtual Client
 Connection ID with the real Client Connection ID before processing the packet further.
 
+## Connection Maintenance in Forwarded Mode
+
+When a client and proxy are using forwarded mode, it is possible that there can be
+long periods of time in which no stream data or datagram frames is exchanged
+between the client and proxy. If these periods extend beyond the max_idle_timeout
+values for the QUIC connection to the proxy (see {{Section 10.1 of QUIC}}),
```
DavidSchinazi, line 578:
```suggestion
values for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
```

draft-pauly-masque-quic-proxy.md
LPardue, line 576:
```suggestion
long periods of time in which no stream data or datagrams are exchanged
```

alternatively maybe just STREAM or DATAGRAM frames

draft-pauly-masque-quic-proxy.md
```diff
@@ -570,6 +570,17 @@ of the received packet to determine if the packet was originated by the proxy,
...
+the QUIC connection might be closed by the proxy if the proxy does not use
+forwarded packets as an explicit liveness signal. To avoid this, clients SHOULD
```
DavidSchinazi, line 580: Why not in the other direction too?

DavidSchinazi (approved):

draft-pauly-masque-quic-proxy.md
LPardue, line 578: Rather than refer to the TP, probably better to mention the effective idle timeout for the connection

draft-pauly-masque-quic-proxy.md
LPardue: I like this as well but was too lazy to offer it as a third suggestion

draft-pauly-masque-quic-proxy.md
tfpauly: I guess the proxy _could_, but if it's not tracking activity, it isn't aware that the client actually isn't idle (and if it is tracking, it doesn't need the keep alive). The client is the only one who knows about all of the connections.

draft-pauly-masque-quic-proxy.md
tfpauly:
```suggestion
which can be done using a PING frame or another ack-eliciting frame as described
in {{Section 10.1.1 of QUIC}}.
```

ehaydenr (approved):

draft-pauly-masque-quic-proxy.md
DavidSchinazi: Can you elaborate? What's preventing the client from timing out?

draft-pauly-masque-quic-proxy.md
```diff
@@ -570,6 +570,18 @@ of the received packet to determine if the packet was originated by the proxy,
 or merely forwarded from the target. The client replaces the Virtual Client
 Connection ID with the real Client Connection ID before processing the packet further.
 
+## Connection Maintenance in Forwarded Mode
+
+When a client and proxy are using forwarded mode, it is possible that there can be
+long periods of time in which no ack-eliciting packets (see {{Section 2 of !QUIC-RETRANSMISSION=RFC9002}}) are exchanged
+between the client and proxy. If these periods extend beyond the max_idle_timeout
+values for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
```
tfpauly, line 578:
```suggestion
between the client and proxy. If these periods extend beyond the effective idle
timeout for the client-to-proxy QUIC connection (see {{Section 10.1 of QUIC}}),
```

draft-pauly-masque-quic-proxy.md
tfpauly: The client is sending forwarded packets and it knows it is sending them actively — if it times out, that is the client's own problem (and it can just send PINGs to keep stuff alive itself).

LPardue (approved):

draft-pauly-masque-quic-proxy.md
tfpauly: Done!

PR #87, opened 2023-10-11
Proposal: Packet Transforms

bemasc:


draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+
+Use of this transform is NOT RECOMMENDED if "scramble" can be deployed.
+
+## "scramble"
+
+The "scramble" transform implements length-preserving unauthenticated encryption
+of QUIC packets while preserving the QUIC invariants.  When "scramble" is in use,
+a global passive adversary cannot select any small set of packets that
+conclusively links the client and target.  However, "scramble" does not defend
+against statistical analysis of large numbers of packets, nor does it protect
```
ehaydenr, line 788: What is considered "large" in this case?

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+When offering or selecting this transform, each party MUST select the key that
+they will use to encrypt scrambled packets and add it to the
+Proxy-QUIC-Transform header in a parameter named "scramble-key" whose value is
+`sf-binary`.
+
+The forward transform operates as follows:
+
+1. Let `k1, k2 = scramble_key[:16], scramble_key[16:32]`.
+1. Instantiate an AES-CTR stream cipher `S` with key `k1`, using a 128-bit counter.
+    * We represent `S` here as a stateful encryption function.
```
ehaydenr, line 801: I _think_ the answer is no, but does any state need to be maintained across packet transforms other than k1 and k2?

bemasc: Correct, packets are processed 100% independently.

draft-ietf-masque-quic-proxy.md
bemasc: I don't have a clear answer for that.  We don't have a very specific model of timing attacks, which are potentially extremely powerful.

I think "scramble" is probably worth defining, but we should certainly take care to avoid over-promising.

draft-ietf-masque-quic-proxy.md
```diff
@@ -98,8 +97,8 @@ proxied packets the target receives.
 target are not able to learn more about the client <-> target communication than
 if no proxy was used.
 
-It is not a goal of forwarded mode to prevent correlation between client <->
-proxy and the proxy <-> target packets from an entity that can observe both
+Forwarded mode does not prevent correlation of client <->
+proxy and proxy <-> target packets by an entity that can observe both
```
tfpauly, line 101: Should we explain a bit more which kinds of correlation are possible here? To explain that trivial correlation isn't possible, but timing/size or active attacks are possible?

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+transforms: "null" and "scramble".
+
+> QUESTION: Should one or both of these be Mandatory To Implement?
+
+## "null"
+
+The "null" transform does not modify the packet in any way.  When this transform
+is in use, a global passive adversary can trivially correlate pairs of packets
+that crossed the forwarder, providing a compact proof that a specific client
+was communicating to a specific target.
```
tfpauly, line 778: Let's add an example of what the whole header looks like

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+   `S.counter = sample`.
+1. Encrypt the first byte: `packet[:1] = S(packet[:1])`.
+1. Restore the Header Form bit: `packet[0] &= 0x7F`.
+1. Encrypt the packet contents after `sample`: `packet[L+17:] = S(packet[L+17:])`.
+1. Encrypt `sample` using a block cipher: `packet[L+1:L+17] = B(sample)`.
+
+NOTE: The security of this arrangement relies on the QUIC payload containing a
+16-byte `sample` that is pseudorandom.  This is guaranteed in QUICv1, but future
+versions of QUIC could in principle produce packets that cannot safely be
+processed by the "scramble" transform.
```
tfpauly, line 816:
```suggestion
processed by the "scramble" transform. As such future versions are defined, new transforms
can be defined to handle forwarding for those packets.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+
+The "scramble" transform implements length-preserving unauthenticated encryption
+of QUIC packets while preserving the QUIC invariants.  When "scramble" is in use,
+a global passive adversary cannot select any small set of packets that
+conclusively links the client and target.  However, "scramble" does not defend
+against statistical analysis of large numbers of packets, nor does it protect
+privacy against an active attacker.
+
+The "scramble" transform is initialized using a 32-byte random symmetric key.
+When offering or selecting this transform, each party MUST select the key that
```
tfpauly, line 792: Let's spell out the fact that both the client and server generate their own key and sends it to the other side

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+they will use to encrypt scrambled packets and add it to the
+Proxy-QUIC-Transform header in a parameter named "scramble-key" whose value is
+`sf-binary`.
+
+The forward transform operates as follows:
+
+1. Let `k1, k2 = scramble_key[:16], scramble_key[16:32]`.
```
tfpauly, line 799: On call, discussed having these be derived off of a common key

draft-ietf-masque-quic-proxy.md
bemasc, line 816:
```suggestion
@DavidSchinazi How about this?

```suggestion
NOTE: The security of this arrangement relies on the 16 bytes following the
Connection ID (i.e., the `sample`) being pseudorandom.  This is guaranteed
in QUICv1, but future versions of QUIC might not maintain this arrangement.
The "scramble" transform MUST NOT be used with any QUIC version that
does not guarantee this.
```

draft-ietf-masque-quic-proxy.md
tfpauly: Although saving the space in the header probably doesn't help. So keep this for now?

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+1. Instantiate an AES-CTR stream cipher `S` with key `k1`, using a 128-bit counter.
```
bemasc, line 800: This syntax is perhaps too customized.  Let's try to stick closer to the style of RFC 9001.

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+1. Instantiate an AES-ECB block cipher `B` with key `k2`.
+1. Let `L` be the Connection ID length.
+1. Let `sample = packet[L+1:L+17]`, i.e., the 16 bytes following the
+   Connection ID.
+1. Set the stream cipher counter to the sample in network byte order:
+   `S.counter = sample`.
+1. Encrypt the first byte: `packet[:1] = S(packet[:1])`.
+1. Restore the Header Form bit: `packet[0] &= 0x7F`.
+1. Encrypt the packet contents after `sample`: `packet[L+17:] = S(packet[L+17:])`.
+1. Encrypt `sample` using a block cipher: `packet[L+1:L+17] = B(sample)`.
```
tfpauly, line 811: Let's also write the same list in reverse order to decrypt

bemasc:
Let's add some examples in this PR.

draft-ietf-masque-quic-proxy.md
bemasc: I've adjusted this text to remove mention of "large numbers" and "statistical analysis", neither of which is required when there are few users of a proxy.

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,6 +743,78 @@ headers does not apply.
...
+of QUIC packets while preserving the QUIC invariants.  When "scramble" is in use,
+a global passive adversary cannot select any small set of packets that
+conclusively links the client and target.  However, "scramble" does not defend
+against statistical analysis of large numbers of packets, nor does it protect
+privacy against an active attacker.
+
+The "scramble" transform is initialized using a 32-byte random symmetric key.
+When offering or selecting this transform, each party MUST select the key that
+they will use to encrypt scrambled packets and add it to the
+Proxy-QUIC-Transform header in a parameter named "scramble-key" whose value is
```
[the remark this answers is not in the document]
bemasc: Done

draft-ietf-masque-quic-proxy.md
bemasc: Done

draft-ietf-masque-quic-proxy.md
bemasc: Done (with some compactification).

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,9 +744,103 @@ headers does not apply.
...
+   modify the Connection ID.
+
+More complex transform behaviors could have internal state, but no such transforms
+are presented here.
+
+Packet transforms are identified by an IANA-registered name, and negotiated in
+the HTTP headers (see {{client-behavior}}).  This document defines two initial
+transforms: "null" and "scramble".
+
+> QUESTION: Should one or both of these be Mandatory To Implement?
```
bemasc, line 772: Let's change this to note that the client doesn't have to use forwarded mode if the proxy chooses a transform they don't like.

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,9 +744,104 @@ headers does not apply.
...
+   (AES-CTR encryption and decryption are identical.)
+
+The encryption keys used in this procedure do not depend on the packet contents,
+so each party only needs to perform AES initialization once for each connection.
+
+NOTE: The security of this arrangement relies on the 16 bytes following the
+Connection ID (i.e., the `sample`) being pseudorandom.  This is guaranteed
+in QUICv1, but future versions of QUIC might not maintain this arrangement.
+The "scramble" transform MUST NOT be used with any QUIC version that
+does not guarantee this property.
```
bemasc, line 839: This needs a little bit more work due to the hypothetical use of CCM_8 in QUICv1.

draft-ietf-masque-quic-proxy.md
bemasc: Done

draft-ietf-masque-quic-proxy.md
bemasc: Done

draft-ietf-masque-quic-proxy.md
```diff
@@ -599,6 +606,13 @@ by including a "Proxy-QUIC-Forwarding" header in a successful response.
 If it supports QUIC packet forwarding, it sets the value to "?1"; otherwise,
 it sets it to "?0".
```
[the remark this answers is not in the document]
tfpauly: The text currently says this:

```
If the client wants to enable QUIC packet forwarding for this request, it sets the value to "?1". If it doesn't want to enable forwarding, but instead only provide information about QUIC Connection IDs for the purpose of allowing the proxy to share a target-facing socket, it sets the value to "?0".
```

So if you don't support any QUIC-aware, you don't include the header
If you support multiplexing only, you send ?0
If you support forwarding, you send ?1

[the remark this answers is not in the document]
bemasc: OK, thanks.  That's the for client, but I see that the proxy-side text is meant to convey the same idea.

draft-ietf-masque-quic-proxy.md
```diff
@@ -734,9 +746,104 @@ headers does not apply.
...
+   Connection ID.  This is the AES-CTR initialization vector and counter.
+1. Let `aesctr0 = AES-ECB(k1, counter)`.  (`len(aesctr0) == 16`.)
+1. Encrypt the lower 7 bits of the first byte:\\
+   `packet[0] ^= aesctr0[0] & 0x7F`.
+    * Here "`^`" represents bitwise XOR and "`&`" represents bitwise AND.
+1. Encrypt up to 15 bytes following the sample:\\
+   `packet[L+17:L+32] ^= aesctr0[1:]`.
+1. Encrypt the remainder of the packet (if any) using AES-CTR:
+    1. Let `i = L + 32`.
+    1. `while i < len(packet)`:
```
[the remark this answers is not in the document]
bemasc: We can, but I haven't found a reference for an AES-CTR primitive so we would still need to define it.  Also, if the primitive "performs encryption" (vs. "provides a keystream") then there would be some extra complexity in explaining how to deal with the first byte.

There's an additional issue that AES-CTR is conventionally defined with an IV and a counter that starts at zero, but here we need the counter to (1) start at some specified value (not zero) and (2) increment as a uint128, not a uint32 or uint64 (due to the potential for the random initial counter value to be near MAX_UINT32 or MAX_UINT64).  (Technically we could also treat it as a uint32 and allow it to wrap on overflow, but a pre-packaged AES-CTR implementation should fail rather than wrap, to avoid the two-time pad problem.)

Anyway, feel free to suggest an alternative formulation if you think you see a clearer way.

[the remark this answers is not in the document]
DavidSchinazi: The first byte issue could be solved by doing something like
```
cleartext = packet[0] | packet[L+17:]`
ciphertext = AES-CTR(plaintext)
encrypted_packet = ciphertext[0] | packet[1:L] | ciphertext[1:]
```
it's not necessarily how this would be implemented, but it's a lot easier to reason about

Can you elaborate on the custom counter needed here? I would assume that we're fine with the counter starting at 0 and the nonce being 16 bytes - in that construction you XOR the nonce and counter to create the IV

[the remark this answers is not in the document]
bemasc: OK, I've updated the PR to pull in a formal definition of AES-CTR from NIST, with appropriate text to specify which variation of AES-CTR we're using, and assign a standard syntax for it.

DavidSchinazi (approved):
Thanks, this is way easier to reason about

draft-ietf-masque-quic-proxy.md
```diff
@@ -454,14 +467,18 @@ The "Proxy-QUIC-Forwarding" is an Item Structured Header {{!RFC8941}}. Its
 value MUST be a Boolean. Its ABNF is:
 
 ~~~
-    Proxy-QUIC-Forwarding = sf-boolean
+    Proxy-QUIC-Forwarding = sf-boolean parameters
```
LPardue, line 470: HTTP editorial style is not to use ABNF, so you can save yourself the effort here and just delete the whole  block

PR #89, opened 2023-10-26
Note that the QUIC bit cannot be used with forwarding mode

bemasc:


LPardue:
You might want to cite or defer to the managability draft that is related - https://www.rfc-editor.org/rfc/rfc9312.html#section-3.1

bemasc:
@LPardue Thanks, that's interesting.  I couldn't figure out how to reference it here though.

It occurs to me that we might also want to say something about disabling the Spin Bit, which is normally a bit dangerous in forwarding mode because it reveals the end-to-end latency of each Connection ID.  However, I'm not sure exactly what the advice would be, especially because it's possibly safe when "scramble" is enabled (#87).

LPardue:
I was thinking something simple like - "The considerations of Identify QUIC traffic in Section 3.1 of QUIC-MANAGEABILTIY] apply to forwarding proxies. Specifically, ... <the text you already have>" 

but if that doesn't help, or feels shoehorned, feel free to ignore

bemasc:
It's tricky because the proxy is both an endpoint and a middlebox (for QUIC).  QUIC-MANAGEABILITY doesn't apply to endpoints.  In the proxy's role as a middlebox, there is no protocol identification involved because (formally) only QUIC can flow through the forwarder, so QUIC-MANAGEABILITY doesn't apply there either.

The weird thing is that the forwarder shares a 5-tuple with the proxy endpoint, so some of the protocol identification restrictions on middleboxes spill over onto the proxy's QUIC endpoint.

ehaydenr:
This lgtm, but since this was opened we added an invariance section here: https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/blob/10e49bcc6ffbc57631752e96cf707233c2909d31/draft-ietf-masque-quic-proxy.md?plain=1#L1467. 

We may want to move the bulk of this to that section and link to it 🤔

Issue #88, opened 2023-10-23
Infinite Loops with Client Virtual Connection IDs


martinduke:
As described in thread that begins in https://mailarchive.ietf.org/arch/msg/masque/y4M_VX7-dMvOqkhKoHCtKUgnomA/

I believe the possible mitigations are:
- Have the server pick the virtual client connection ID
- Require the proxy to open a new target-facing UDP socket for each connection request
- Not do virtual client connection IDs at all

tfpauly:
Or require that if servers share a VIP, they need to coordinate on their VCID mappings? That seems to be a simple requirement.

I don't think removing VCIDs or allowing servers to select the CIDs that clients advertise to targets is feasible.

martinduke:
> Or require that if servers share a VIP, they need to coordinate on their VCID mappings? That seems to be a simple requirement.

It sounds very expensive to globally share all of the VCIDs that clients have chosen!

> 
> I don't think removing VCIDs or allowing servers to select the CIDs that clients advertise to targets is feasible.

Why? The client sends REGISTER_CLIENT_CID to the proxy, but without including a virtual. The client can't start sending packets with that CID because the proxy hasn't confirmed that there is no collision at its target-facing socket. The ACK_CLIENT_CID could contain the Virtual client CID the proxy wants the client to use.

This is, IIUC, exactly how target CIDs work.

tfpauly:
I think it's OK if we wanted the VCID to be determined by the proxy — I was saying that it's not OK for the actual client CID to be determined by the proxy.

If we just have the VCID be determined by the proxy, does that sufficiently solve the issue? If so, that's probably the best outcome.

martinduke:
If the proxy picks the virtual CID, it also MUST NOT just reuse the client-provided one, because of this vulnerability.

The same goes for target connection IDs: I can easily recreate this attack in the forward direction if the proxy just reuses the target CID for the virtual. In fact, if the relationship between CID and Virtual CIDs is **predictable**, I can do a forward direction infinite loop with a colluding target.

But I guess that's already good advice if we're trying to prevent linkabilty between target-side and client-side.

tfpauly:
Should be fixed with #104

PR #104, opened 2024-02-19
Proxy-Chosen Virtual Client Connection ID

ehaydenr:
As described in #88, loop attacks are possible when clients pick the virtual connection ID. This change moves the responsibility of generating a Virtual Client Connection ID to the proxy and requires the proxy to generate unpredictable virtual connection IDs.

Unfortunately, this change complicates the capsule exchange. Specifically, the proxy cannot send forwarded mode packets in the Target->Client direction until it knows that the client is ready to receive them. Previously, when the client chose the vcid, we could require that the client not share the vcid unless it's ready to receive with it. Now that the proxy chooses the client vcid, we need the client to signal it's ready to receive forwarded mode packets. To accomplish this, ACK_CLIENT_VCID is introduced.

The ACK_CLIENT_VCID capsule solves the rule-readiness problem and maintains that the client can supply a Stateless Reset Token for resetting the client<->target tunnel.

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -193,7 +193,7 @@ ID. Similarly, clients multiplexing connections on the same UDP socket SHOULD
 choose a Client Connection ID that's sufficiently long to reduce the likelihood
 of a conflict with the proxy-chosen Virtual Client Connection ID. The Virtual
 Client Connection ID MUST either be constructed such that it is unpredictable to
-the client or to guarantee no conflicts among all proxyies sharing an IP address
+the client or to guarantee no conflicts among all proxies sharing an IP address
 and port. See {{security}} for more discussion on Virtual Client Connection ID
 construction.
 
@@ -1018,7 +1018,6 @@ attacks. Preventing Virtual Client Connection ID conflicts across proxies
 sharing an IP address and port mitigates one such forwarding loop attack.
 Conflicts can be avoided by partitioning the Virtual Client Connection ID space
 across proxies, using sufficiently long and random values, or by other means.
-
 
 [comment1]: # OPEN ISSUE: Figure out how clients and proxies could interact to
 [comment2]: # learn whether an adversary is injecting malicious forwarded
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -189,7 +189,13 @@ Client Connection ID allows the connection ID bytes to change on the wire
 without requiring the connection IDs on the client to target connection change.
 To reduce the likelihood of connection ID conflicts, the proxy SHOULD choose a
 Virtual Client Connection ID that is at least as long as the Client Connection
-ID.
+ID. Similarly, clients multiplexing connections on the same UDP socket SHOULD
+choose a Client Connection ID that's sufficiently long to reduce the likelihood
+of a conflict with the proxy-chosen Virtual Client Connection ID. The Virtual
+Client Connection ID MUST either be constructed such that it is unpredictable to
+the client or to guarantee no conflicts among all proxyies sharing an IP address
+and port. See {{security}} for more discussion on Virtual Client Connection ID
+construction.
 
 Clients and Proxies not implementing forwarded mode do not need to consider
 Virtual Connection IDs since all Client<->Target datagrams will be encapsulated
@@ -1006,6 +1012,13 @@ Since proxies that forward QUIC packets do not perform any cryptographic
 integrity check, it is possible that these packets are either malformed,
 replays, or otherwise malicious. This may result in proxy targets rate limiting
 or decreasing the reputation of a given proxy.
+
+Chaining of proxies using forwarded mode introduces the risk of forwarding loop
+attacks. Preventing Virtual Client Connection ID conflicts across proxies
+sharing an IP address and port mitigates one such forwarding loop attack.
+Conflicts can be avoided by partitioning the Virtual Client Connection ID space
+across proxies, using sufficiently long and random values, or by other means.
+
 
 [comment1]: # OPEN ISSUE: Figure out how clients and proxies could interact to
 [comment2]: # learn whether an adversary is injecting malicious forwarded
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -161,14 +161,13 @@ the client.
 is used in the Destination Connection ID field of packets from the client to
 the target.
 - Virtual Client Connection ID: a fake QUIC Connection ID that is chosen by the
-client that the proxy MUST use when sending QUIC packets in forwarded mode.
+proxy that the proxy MUST use when sending QUIC packets in forwarded mode.
 - Virtual Target Connection ID: a fake QUIC Connection ID that is chosen by the
 proxy that the client MUST use when sending QUIC packets in forwarded mode.
 
 ## Virtual Connection IDs
 
-QUIC allows each endpoint of a connection to choose the connection IDs it
-receives with. Servers receiving QUIC packets can employ load balancing
+Servers receiving QUIC packets can employ load balancing
 strategies such as those described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}
 that encode routing information in
 the connection ID. When operating in forwarded mode, clients send QUIC packets
@@ -182,7 +181,7 @@ forwarding the packet to the Target.
 
 Similarly, QUIC requires that connection IDs aren't reused over multiple network
 paths to avoid linkability. The Virtual Client Connection ID is a connection ID
-chosen by the Client that the Proxy uses when sending forwarded mode packets.
+chosen by the Proxy that the Proxy uses when sending forwarded mode packets.
 The Proxy replaces the Client Connection ID with the Virtual Client Connection
 ID prior to forwarding the packet to the Client. Clients take advantage of this
 to avoid linkability when migrating a client to proxy network path. The Virtual
@@ -339,8 +338,8 @@ using HTTP Datagrams with the context ID set to zero as defined in
 {{CONNECT-UDP}}.
 
 The capsules used for QUIC-aware proxying allow a client to register connection
-IDs with the proxy, and for the proxy to acknowledge or reject the connection
-ID mappings.
+IDs with the proxy, and for the client and proxy to acknowledge or reject the
+connection ID mappings.
 
 The REGISTER_CLIENT_CID and REGISTER_TARGET_CID capsule types (see
 {{iana-capsule-types}} for the capsule type values) allow a client to inform
@@ -350,8 +349,17 @@ respectively. These capsule types MUST only be sent by a client.
 The ACK_CLIENT_CID and ACK_TARGET_CID capsule types (see {{iana-capsule-types}}
 for the capsule type values) are sent by the proxy to the client to indicate
 that a mapping was successfully created for a registered connection ID as well
-as provide the Virtual Target Connection ID that may be used in forwarded mode.
+as provide the Virtual Connection IDs that may be used in forwarded mode.
 These capsule types MUST only be sent by a proxy.
+
+The ACK_CLIENT_VCID capsule type MUST only be sent by the client and only when
+forwarding mode is enabled. It is sent by the client to the proxy in response to
+an ACK_CLIENT_CID capsule to indicate that the client is ready to receive
+forwarded mode packets with the specified virtual connection ID. The proxy MUST
+NOT send forwarded mode packets to the client prior to receiving this
+acknowledgement. This capsule also contains a Stateless Reset Token the client
+may respond with when receiving forwarded mode packets with the specified
+virtual connection ID.
 
 The CLOSE_CLIENT_CID and CLOSE_TARGET_CID capsule types (see
 {{iana-capsule-types}} for the capsule type values) allow either a client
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -918,12 +1029,13 @@ registry established by {{HTTP-DGRAM}}.
 
 |     Capule Type     |   Value   | Specification |
 |:--------------------|:----------|:--------------|
-| REGISTER_CLIENT_CID | 0xffe400  | This Document |
-| REGISTER_TARGET_CID | 0xffe401  | This Document |
-| ACK_CLIENT_CID      | 0xffe402  | This Document |
-| ACK_TARGET_CID      | 0xffe403  | This Document |
-| CLOSE_CLIENT_CID    | 0xffe404  | This Document |
-| CLOSE_TARGET_CID    | 0xffe405  | This Document |
+| REGISTER_CLIENT_CID | 0xffe500  | This Document |
```
tfpauly, line 1032: Thanks for updating the values!

DavidSchinazi:
If the virtual client CID is chosen by the proxy, doesn't that mean that the client can no longer multiplex multiple connections on the same socket?

ehaydenr:
> If the virtual client CID is chosen by the proxy, doesn't that mean that the client can no longer multiplex multiple connections on the same socket?

I don't think so. The difference here is that the client has to create a demultiplexing rule with a value that the proxy chooses as opposed to one it chose itself. And the proxy can't send forwarded mode packets until the client signals that the rule is created.

Maybe I'm missing something, can you elaborate?

DavidSchinazi:
In general the mindset of QUIC CIDs is that they're picked by the receiver. That allows among other things the receiver to pick their CID length (a length 1 might be sufficient for a client that's only multiplexing two connections, but a server multiplexing hundreds would want more bytes). If you let someone else pick the CIDs, you could end up with a conflict on the client if it's talking to two proxies and the first proxy picks `CID=12345678` and the other picks `CID=1234567890`. One potential solution is to require a sufficient length that makes conflicts statistically unlikely, but the min length depends on how much multiplexing the client wants, so now maybe we need a way to encode that length?

ehaydenr:
@DavidSchinazi what do you think about 98b876ef2ff61d02ec91722bd754c7e49ad696ad?

DavidSchinazi:
I like that. We'll probably want to add some implementation considerations that
* tell proxies that they need to select these CIDs randomly (maybe even MUST?)
* tell clients that if they're multiplexing they need to pick a client connection ID length that makes conflicts unlikely if the proxy picks a random number for them

but in general this should work

ehaydenr:
@DavidSchinazi, how does the language added in 4a2dc2164b69149ee54b2c76678d6791847907c0 sound?

DavidSchinazi (approved):
Thanks for the latest updates, I think this is good

draft-ietf-masque-quic-proxy.md
```diff
@@ -182,12 +181,21 @@ forwarding the packet to the Target.
...
 to avoid linkability when migrating a client to proxy network path. The Virtual
 Client Connection ID allows the connection ID bytes to change on the wire
 without requiring the connection IDs on the client to target connection change.
+To reduce the likelihood of connection ID conflicts, the proxy SHOULD choose a
+Virtual Client Connection ID that is at least as long as the Client Connection
+ID. Similarly, clients multiplexing connections on the same UDP socket SHOULD
+choose a Client Connection ID that's sufficiently long to reduce the likelihood
+of a conflict with the proxy-chosen Virtual Client Connection ID. The Virtual
+Client Connection ID MUST either be constructed such that it is unpredictable to
+the client or to guarantee no conflicts among all proxyies sharing an IP address
```
tfpauly, line 196:
```suggestion
the client or to guarantee no conflicts among all proxies sharing an IP address
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -891,6 +1013,13 @@ integrity check, it is possible that these packets are either malformed,
 replays, or otherwise malicious. This may result in proxy targets rate limiting
 or decreasing the reputation of a given proxy.
 
+Chaining of proxies using forwarded mode introduces the risk of forwarding loop
+attacks. Preventing Virtual Client Connection ID conflicts across proxies
+sharing an IP address and port mitigates one such forwarding loop attack.
+Conflicts can be avoided by partitioning the Virtual Client Connection ID space
+across proxies, using sufficiently long and random values, or by other means.
+
+
```
tfpauly, line 1022:
```suggestion
```

tfpauly (approved):

marten-seemann:
I have two concerns with this approach:
1. Unless I'm missing something, this design assumes that there's no CID-based routing happening inside the client's network. I don't feel comfortable hard-coding this assumption into the design here.
2. Similar to @DavidSchinazi, I'm concerned about the CID length here. My QUIC implementation assumes that all CIDs have the same (configurable) length. When a new packet arrives, I read `n` CID bytes, and find the respective QUIC connection by doing a map lookup. This doesn't work anymore if the proxy gives me a CID that has a different length. I'd either have to build some kind of prefix-trie, or do (up to `20-n`) map lookups. Given that the CID lookup already shows up in CPU profiles during high-bandwidth transfers, both of these options seem suboptimal. Things get even more complicated if an implementation decides to encode the CID length into the CID itself.

ehaydenr:
@marten-seemann, to address your second point, one option could be to change this from a SHOULD to a MUST
> To reduce the likelihood of connection ID conflicts, the proxy SHOULD choose a
 Virtual Client Connection ID that is at least as long as the Client Connection

Your first point is well taken. For what it's worth, this requirement mirrors the requirement for the proxy - it must receive QUIC packets (with CIDs it did not choose) from the target and demultiplex them.

As mentioned at 119, I see three options for addressing the loop issue:
1. Prohibit VIP sharing
2. Eliminate target port sharing
3. Proxy-chosen Virtual Client CIDs (this PR)

While certainly not perfect, I think this option is the best of those three. Do you disagree? Do you see a another alternative?

martinduke:
> I have two concerns with this approach:
> 
> 1. Unless I'm missing something, this design assumes that there's no CID-based routing happening inside the client's network. I don't feel comfortable hard-coding this assumption into the design here.

This is already a property of the design, before this PR, unless I'm missing something. Proxies are not able to choose the client CID that the target writes into packets, except for an ability to veto ones that collide. There is no ability to propose a routable connection ID for use on the target-facing interface.

ehaydenr:
@martinduke, if I'm understanding @marten-seemann correctly, I think he's referring to the fact that, with this change, the client can no longer choose the virtual client connection IDs that it receives with in forwarding mode. I do agree with you that the proxy has never been able to influence the _real_ client connection IDs beyond rejecting collisions - and this PR doesn't change that.

martinduke:
> @martinduke, if I'm understanding @marten-seemann correctly, I think he's referring to the fact that, with this change, the client can no longer choose the virtual client connection IDs that it receives with in forwarding mode. I do agree with you that the proxy has never been able to influence the _real_ client connection IDs beyond rejecting collisions - and this PR doesn't change that.

I agree with your statement of the facts. But my point is that we've already negated the idea of connection-ID based load balancing on the client side. Since QUIC (modulo MPQUIC) doesn't allow changes to server address anyway, this is not much of a problem.

ehaydenr:
@marten-seemann, thoughts?

marten-seemann:
Sorry for the late response, I'll get back to this later today or tomorrow.

marten-seemann:
> 2\. Similar to @DavidSchinazi, I'm concerned about the CID length here. My QUIC implementation assumes that all CIDs have the same (configurable) length. When a new packet arrives, I read `n` CID bytes, and find the respective QUIC connection by doing a map lookup. This doesn't work anymore if the proxy gives me a CID that has a different length. I'd either have to build some kind of prefix-trie, or do (up to `20-n`) map lookups. Given that the CID lookup already shows up in CPU profiles during high-bandwidth transfers, both of these options seem suboptimal. Things get even more complicated if an implementation decides to encode the CID length into the CID itself.

I spent some more time thinking about this point, and I'm wondering if I'm missing something here. Compared to a simple map lookup, looking up a variable-length CID is way slower. I [implemented](https://github.com/marten-seemann/cid-lookup/blob/master/trie.go) a prefix trie with (up to) 20 levels, and its performance is terrible. At 100k CIDs, a CID lookup takes up to 40x longer:
```
BenchmarkTrieLookup/4_byte_CIDs-16              20221346                64.73 ns/op
BenchmarkTrieLookup/8_byte_CIDs-16               5931970               207.9 ns/op
BenchmarkTrieLookup/15_byte_CIDs-16              2255654               523.1 ns/op
BenchmarkTrieLookup/20_byte_CIDs-16              1402430               805.1 ns/op
BenchmarkMapLookup/4_byte_CIDs-16               88730054                13.92 ns/op
BenchmarkMapLookup/8_byte_CIDs-16               91244919                13.14 ns/op
BenchmarkMapLookup/15_byte_CIDs-16              85435095                13.84 ns/op
BenchmarkMapLookup/20_byte_CIDs-16              41338683                29.12 ns/op
```

In these cases, it would be faster to do a map lookup for all 20 possible CID lengths, until a match is found. 

Admittedly, this problem exists on the proxy side already, but now we're moving it to the client as well. There are probably faster algorithms than a prefix trie (prefix b-trees), but I'd be curious to learn what kind of data structure people are planning to use for this in practice.

marten-seemann:
I'm not sure if this is a nice solution, but an alternative could be using a bloom filter. I described the idea in more detail in https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/issues/108.

afressancourt:
Would it be possible to let the client choose between the method initially described in the draft and the mechanism proposed by @ehaydenr ? 
In the signaling, if the REGISTER_CLIEND_ID contains only a Client Connection ID, then we are letting the proxy propose the virtual Client CID, and if the REGISTER_CLIEND_ID contains both a Client Connection ID and a virtual Client Connection ID, the proxy tries to accommodate this request, and sends an error message if it is not possible to deal with a client-set virtual Client Connection ID ?

martinduke:
> Would it be possible to let the client choose between the method initially described in the draft and the mechanism proposed by @ehaydenr ? In the signaling, if the REGISTER_CLIEND_ID contains only a Client Connection ID, then we are letting the proxy propose the virtual Client CID, and if the REGISTER_CLIEND_ID contains both a Client Connection ID and a virtual Client Connection ID, the proxy tries to accommodate this request, and sends an error message if it is not possible to deal with a client-set virtual Client Connection ID ?

If I'm not mistaken, if the proxy allows the client to choose the connection ID, it is open to the attack that is the reason for this PR.

afressancourt:
> If I'm not mistaken, if the proxy allows the client to choose the connection ID, it is open to the attack that is the reason for this PR.

I understand the wish to mitigate loop attacks, but at the same time it complicates the connection setup with one additional message. Besides, we can argue that not answering the ACK_CLIENT_ID sent by the proxy can be used as an attack vector if clients coordinate to sent a high number of REGISTER_CLIENT_CID to the same proxy. 

Letting the client choose whether he wants to let the proxy choose its virtual connection ID would allow to either ease the capsule exchange at connection setup, or fallback in a mode which protects the proxy against loop attacks.

ehaydenr:
> I understand the wish to mitigate loop attacks, but at the same time it complicates the connection setup with one additional message.

I would argue the additional message is cheap since it requires no additional round trip.

> Besides, we can argue that not answering the ACK_CLIENT_ID sent by the proxy can be used as an attack vector if clients coordinate to sent a high number of REGISTER_CLIENT_CID to the same proxy.

The proxy can reject the CID registration with the corresponding CLOSE_CLIENT_CID/CLOSE_TARGET_CID capsule. For adding guidance here, I created #109 

> Letting the client choose whether he wants to let the proxy choose its virtual connection ID would allow to either ease the capsule exchange at connection setup, or fallback in a mode which protects the proxy against loop attacks.

Are you suggesting we keep the capsule exchange as-is and for deployments that wish to mitigate loop attacks, they simply don't implement forwarding mode?

ehaydenr:
@marten-seemann , does my comment about skipping variable length lookups address your concern?

marten-seemann:
> @marten-seemann, one way of avoiding variable length lookups is to use a minimum client cid size and reject conflicts. Let's say you want to use N-byte CIDs. You would send a REGISTER_CLIENT_CID capsule with your N-byte CID and receive a virtual client connection ID back that's N+M bytes long. For the purposes of connection lookup, just use the first N bytes. If the first N bytes conflicts with some other connection, don't proceed with forwarding mode or send another REGISTER_CLIENT_CID capsule to solicit a new virtual client connection ID.

This makes sense to me. It assumes that there's not too much structure to CIDs issued (e.g. if the proxy's CID allocation strategy would be to use an N byte prefix, this method wouldn't work), but I think that's guaranteed by the fact that CIDs MUST be unlinkable to an on-path observer.

afressancourt:
> > I understand the wish to mitigate loop attacks, but at the same time it complicates the connection setup with one additional message.
> 
> I would argue the additional message is cheap since it requires no additional round trip.

If you can send the ACK_CLIEN_VCID together with the first bytes of data, then, I agree, you "only" loose the overhead of the ACK message. 

> > Letting the client choose whether he wants to let the proxy choose its virtual connection ID would allow to either ease the capsule exchange at connection setup, or fallback in a mode which protects the proxy against loop attacks.
> 
> Are you suggesting we keep the capsule exchange as-is and for deployments that wish to mitigate loop attacks, they simply don't implement forwarding mode?

Not exactly. I had in mind a mechanism in which the client starts with sending a REGISTER_CLIENT_ID with either a VCCID or not in it (or a bit stating the VCCID is proposed but can be overridden). 
If the proxy receives a REGISTER_CLIENT_ID with either no VCCID or an overriddable VCCID, then:
- It can propose a VCCID in the ACK_CLIENT_ID, which will be acked itself by the ACK_CLIENT_VCID. 
- Otherwise, we are in a state in which the client does want to use its VCCID, so then either the proxy acks this sending a ACK_CLIENT_ID or it refuses the connection with the corresponding CLOSE_CLIENT_CID/CLOSE_TARGET_CID capsule.

PR #111, opened 2024-07-03
Full editorial pass

tfpauly:
Major editorial pass to re-structure the document for clarity, flow, etc. All of the old content is there, but moved around, and this tries to explain things better in the overview section.

Rendered version:
https://ietf-wg-masque.github.io/draft-ietf-masque-quic-proxy/tfp/full-editorial/draft-ietf-masque-quic-proxy.html

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -307,7 +307,7 @@ the mappings between VCIDs, CIDs, and 4-tuples are unique. Specifically,
 in order to route packets sent by the client, the proxy needs to be able
 to observe the VCID and the client-to-proxy 4-tuple, and map them
 to a specific target CID and proxy-to-target 4-tuple. In order to route
-packets send by a target, the proxy needs to be able to observe the client
+packets sent by a target, the proxy needs to be able to observe the client
 CID and the proxy-to-target 4-tuple, and map them to a specific VCID
 and client-to-proxy 4-tuple. Since proxies choose the VCID values, they
 can ensure that the VCIDs are distinguishable.
@@ -317,8 +317,7 @@ strategies such as those described in {{?QUIC-LB=I-D.ietf-quic-load-balancers}}
 that encode routing information in the connection ID. When operating in
 forwarded mode, clients send QUIC packets destined for the target directly
 to the proxy. Since these packets are generated using the target CID,
-load balancers would not be able to route packets to the correct proxy if the
-packets were sent with the target CID. The target VCID
+load balancers may not have the necessary information to route packets to the correct proxy. The target VCID
 is a VCID chosen by the proxy that the client uses when sending
 forwarded mode packets. The proxy replaces the target VCID
 with the target CID prior to forwarding the packet to the target.
@@ -457,8 +456,8 @@ set the header field value to "?0", or omit the header entirely.
 
 Connection ID awareness relies on using capsules {{HTTP-CAPSULES}} to
 signal addition and removal of Connection IDs. Clients send capsules
-to let proxies know when new Connection IDs on the client-to-target
-QUIC connection are changed. Proxies send capsules to acknowledge or
+to let proxies know when Connection IDs on the client-to-target
+QUIC connection are changing. Proxies send capsules to acknowledge or
 reject these Connection IDs, and in forwarded mode to let clients know
 about Virtual Connection IDs to use on the client-to-proxy link.
 
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -197,15 +197,15 @@ a main motivation for enabling forwarded mode), each subsequent proxy
 is just treated as a target by the preceding proxy.
 
 ~~~ aasvg
-   +--------+              +-----------+              +--------+   
-   |        |              |           |              |        |   
-   | Client +--------------+   Proxy   +--------------+ Target |   
-   |        |              |           |              |        |   
-   +--------+              +-----------+              +--------+  
+   +--------+              +-----------+              +--------+
+   |        |              |           |              |        |
+   | Client +--------------+   Proxy   +--------------+ Target |
+   |        |              |           |              |        |
+   +--------+              +-----------+              +--------+
      
        <----------- Client-to-target connection ----------->
-       
-       
+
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
+              connection
+
+       CCCC1111                                     1111CCCC
+       <----------- Client-to-target connection ----------->
+
+  AAAABBBB (CCCC1111)   BBBBAAAA (1111CCCC)
+       <-- Client-to-proxy -->
+                VCIDs
+~~~
+{: #vcid-diagram title="Diagram of VCIDs in forwarded mode"}
```
ehaydenr, line 302: It seems like we're jumping between a few different conventions for CIDs and I'm wondering if there's an opportunity for more consistency.

In the "Example of sharing a proxy-to-target 4-tuple" diagram prior to this, I see the following for the client<->target connection:
- 4 of the same letter (the client's letter) followed by the same 4 numbers for client CID
- Those same 4 letters followed by those 4 numbers all incremented by one

In this diagram, we have to show CIDs for three different connections which requires 6 different CIDs. The convention switches to doing 4 letters followed by 4 numbers vs. those same 4 numbers followed by four letters.  I think having a distinct pattern for each of these that's consistent across diagrams might be useful.

I think the CCCC0000, 0000CCCC CID pair pattern may be the least distracting. Each time there's a new client, go to the next letter in the alphabet. Each time a single client needs another CID pair, increment the number. For this diagram, that would mean instead of AAAABBBB we would have CCCC2222

draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
+
+In order for a proxy to correctly route packets using VCIDs from
+client-to-target and target-to-client, the proxy MUST guarantee that
+the mappings between VCIDs, CIDs, and 4-tuples are unique. Specifically,
+in order to route packets sent by the client, the proxy needs to be able
+to observe the VCID and the client-to-proxy 4-tuple, and map them
+to a specific target CID and proxy-to-target 4-tuple. In order to route
+packets send by a target, the proxy needs to be able to observe the client
```
ehaydenr, line 310:
```suggestion
packets sent by a target, the proxy needs to be able to observe the client
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
-to the correct Proxy if the packets were sent with the Target Connection ID.
-The Virtual Target Connection ID is a connection ID chosen
-by the Proxy that the Client uses when sending forwarded mode packets. The Proxy
-replaces the Virtual Target Connection ID with the Target Connection ID prior to
-forwarding the packet to the Target.
+that encode routing information in the connection ID. When operating in
+forwarded mode, clients send QUIC packets destined for the target directly
+to the proxy. Since these packets are generated using the target CID,
+load balancers would not be able to route packets to the correct proxy if the
+packets were sent with the target CID. The target VCID
```
ehaydenr, line 321:
```suggestion
to the proxy. Since these packets are generated using the target CID,
load balancers may not have the necessary information to route packets to the correct proxy. The target VCID
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
 
-Each pair of Virtual Target Connection ID and client-facing socket MUST map to a
-single target-facing socket and Target Connection ID.
+For QUIC-aware proxying without forwarded mode, the steps are as follows:
 
-~~~
-(Client-facing socket + Virtual Target Connection ID)
-    => (Target-facing socket + Target Connection ID)
-~~~
+1. The client sends the `REGISTER_CLIENT_CID` capsule as soon as it starts
```
ehaydenr, line 377: This numbered list all starts with `1.`

draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
+sends a `CLOSE_CLIENT_CID` or `CLOSE_TARGET_CID` capsule. The client
+can also initiate new `REGISTER_CLIENT_CID` or `REGISTER_TARGET_CID`
+exchanges at any time.
 
-~~~
-(Target-facing socket + Client Connection ID) => (Client <-> Proxy HTTP Stream)
-(Target-facing socket + Client Connection ID)
-    => (Client-facing socket + Virtual Client Connection ID)
-~~~
+For QUIC-aware proxying with forwarded mode, the steps are as follows:
```
ehaydenr, line 396: The steps are nearly identical between the two modes - should we consolidate?

draft-ietf-masque-quic-proxy.md
```diff
@@ -168,268 +168,363 @@ This document uses the following terms:
...
 
-# Connection ID Capsule Types
+# Connection ID Capsules {#cid-capsules}
 
-Proxy awareness of QUIC Connection IDs relies on using capsules ({{HTTP-DGRAM}})
-to signal the addition and removal of Client and Target Connection IDs.
+Connection ID awareness relies on using capsules {{HTTP-CAPSULES}} to
+signal addition and removal of Connection IDs. Clients send capsules
+to let proxies know when new Connection IDs on the client-to-target
+QUIC connection are changed. Proxies send capsules to acknowledge or
```
ehaydenr, line 461:
```suggestion
to let proxies know when Connection IDs on the client-to-target
QUIC connection are changing. Proxies send capsules to acknowledge or
```

draft-ietf-masque-quic-proxy.md
tfpauly: Yes that's how markdown numbered lists work, it's weird

draft-ietf-masque-quic-proxy.md
tfpauly: Updated!

draft-ietf-masque-quic-proxy.md
tfpauly: They're close, but there are differences in almost all steps for VCID vs no VCID

PR #118, opened 2024-10-03
MAX_CONNECTION_IDS capsule

### LINKED ISSUE #109: guidance for proxies to limit the number of concurrently registered CIDs
A client can send many bogus REGISTER_TARGET_CID/REGISTER_CLIENT_CID capsules. A proxy has to maintain the forwarding / port demultiplexing rules for each of the registered CIDs it acknowledges. Should proxies signal some limit? Or should they merely CLOSE_CLENT_CID/CLOSE_TARGET_CID when the limit is exceeded?

ehaydenr:
closes #109

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -478,8 +478,9 @@ The REGISTER_CLIENT_CID ({{capsule-reg-client}}) and REGISTER_TARGET_CID
 the proxy about a new client CID or a new target CID,
 respectively. These capsule types MUST only be sent by a client. These capsule
 types share a sequence number space which allows the proxy to limit the
-number of active registrations. The first registration has sequence number 0
-and subsequent registrations increment the sequence number by 1.
+number of active registrations. The first registration (of either client CID or target CID)
+has sequence number 0, and subsequent registrations increment the sequence number
+by 1.
 
 The ACK_CLIENT_CID ({{capsule-ack-client}}) and ACK_TARGET_CID
 ({{capsule-ack-target}}) capsule types are sent by the proxy to the client
@@ -797,7 +798,7 @@ MAX_CONNECTION_IDS capsule. The initial MAX_CONNECTION_IDS value is 1, allowing 
 sequence numbers 0 and 1 for a total of two registrations without receiving a
 MAX_CONNECTION_IDS capsule from the proxy.
 
-Clients which cannot register new connection IDs within a reasonable timeout due to
+Clients that cannot register new connection IDs within a reasonable time due to
 the MAX_CONNECTION_IDS limit SHOULD abort the proxied connection by resetting the HTTP
 stream with error code NO_ERROR. This may happen, for example, if the target server
 sends a NEW_CONNECTION_ID frame with Sequence Number and Retire Prior To equal to the
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -387,6 +387,10 @@ the target CID on the client-to-target connection.
 with no associated target VCID; alternatively, the proxy can send the
 `CLOSE_TARGET_CID` if it detects a conflict with another CID.
 
+1. The proxy sends the `MAX_CONNECTION_IDS` capsule to allow additional
+registration of new connection IDs via future `REGISTER_CLIENT_CID` and
+`REGISTER_TARGET_CID` capsules.
+
 1. Whenever a client or target stops uses a particular CID, the client
 sends a `CLOSE_CLIENT_CID` or `CLOSE_TARGET_CID` capsule. The client
 can also initiate new `REGISTER_CLIENT_CID` or `REGISTER_TARGET_CID`
@@ -412,6 +416,10 @@ with a target VCID; alternatively, the proxy can send the
 `CLOSE_TARGET_CID` if it detects a conflict with another CID. Once
 the client receives the target VCID, it can start sending forwarded
 packets using the target VCID.
+
+1. The proxy sends the `MAX_CONNECTION_IDS` capsule to allow additional
+registration of new connection IDs via future `REGISTER_CLIENT_CID` and
+`REGISTER_TARGET_CID` capsules.
 
 1. Whenever a client or target stops uses a particular CID, the client
 sends a `CLOSE_CLIENT_CID` or `CLOSE_TARGET_CID` capsule. The client
@@ -468,7 +476,10 @@ using HTTP Datagrams with the context ID set to zero as defined in
 The REGISTER_CLIENT_CID ({{capsule-reg-client}}) and REGISTER_TARGET_CID
 ({{capsule-reg-target}}) capsule types allow a client to inform
 the proxy about a new client CID or a new target CID,
-respectively. These capsule types MUST only be sent by a client.
+respectively. These capsule types MUST only be sent by a client. These capsule
+types share a sequence number space which allows the proxy to limit the
+number of active registrations. The first registration has sequence number 0
+and subsequent registrations increment the sequence number by 1.
 
 The ACK_CLIENT_CID ({{capsule-ack-client}}) and ACK_TARGET_CID
 ({{capsule-ack-target}}) capsule types are sent by the proxy to the client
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -495,21 +506,40 @@ CLOSE_CLIENT_CID without having sent an ACK_CLIENT_VCID capsule, the client is
...
 needs the connection ID registered.
 
+The MAX_CONNECTION_IDS capsule type {{capsule-max-cids}} MUST only be sent by the
+proxy. It indicates to the client the maximum permitted sequence number for
+connection ID registrations. This allows the proxy to limit the number of active
+registrations. The initial maximum is 1, allowing the client to send 2 registrations,
+one with sequence number 0 and another with sequence number 1. A proxy MUST NOT
+send a MAX_CONNECTION_IDS capsule with a value less than 1. Clients receiving a
+MAX_CONNECTION_IDS capsule with a value less than 1 MUST reset the stream with
+TODO error code.
```
ehaydenr, line 516: Not sure what stream reset code is most appropriate here 🤔

DavidSchinazi: [H3_DATAGRAM_ERROR](https://www.rfc-editor.org/rfc/rfc9297#name-http-3-error-code)

draft-ietf-masque-quic-proxy.md
```diff
@@ -495,21 +506,40 @@ CLOSE_CLIENT_CID without having sent an ACK_CLIENT_VCID capsule, the client is
...
+  Type (i) = 0xffe600
   Length (i),
+  Sequence Number (i),
+  Connection ID Length (i)
   Connection ID (0..2040),
 }
 ~~~
 {: #fig-capsule-register-client-cid title="Register CID Capsule Format"}
 
+Sequence Number
```
DavidSchinazi, line 535: Do we need this field to be in the capsule? QUIC uses sequence numbers because packet reordering can cause reordering of connection ID frames. Here, on the other hand, we're on the capsule stream and that's ordered, so we can know the sequence number implicitly. I'd suggest removing this field to keep the implementation simpler.

draft-ietf-masque-quic-proxy.md
```diff
@@ -870,6 +922,10 @@ short header packets (based on the Header Form bit) and have mapped target VCIDs
 Packets sent by the client that are forwarded SHOULD be
 considered as activity for restarting QUIC's Idle Timeout {{QUIC}}.
 
+In order to permit the client to change client-to-target connection IDs, the proxy
+SHOULD send MAX_CONNECTION_IDS capsules allowing the client additional connection ID
+registrations.
+
```
DavidSchinazi, line 928: Should we add some guidance for clients as to what they should do if they run out of registrations?

draft-ietf-masque-quic-proxy.md
```diff
@@ -690,13 +724,13 @@ in response to target-to-client forwarded mode packets.
...
-CLOSE_CLIENT_CID and CLOSE_TARGET_CID capsule types have types `0xffe505` and
-`0xffe506`, respectively, and include a single connection ID to close. They
+CLOSE_CLIENT_CID and CLOSE_TARGET_CID capsule types have types `0xffe605` and
+`0xffe606`, respectively, and include a single connection ID to close. They
 can be sent by either clients or proxies.
 
 ~~~
 Close CID Capsule {
-  Type (i) = 0xffe505, 0xffe506
+  Type (i) = 0xffe605, 0xffe606
```
DavidSchinazi, line 733: Pro-tip: in previous drafts I would write here `see {{iana}} for the value of the capsule type` so that we only need to modify a single value when we rev the draft

draft-ietf-masque-quic-proxy.md
ehaydenr: Perhaps the is _most_ appropriate, but it feels a little odd to me since the description is "Datagram or Capsule Protocol parse error". We're not failing to parse the capsule protocol, we're just unhappy with a value that happened to be delivered over a capsule.

draft-ietf-masque-quic-proxy.md
ehaydenr: It's not needed, no. I don't have a strong opinion, but the advantage I see is it may make it easier to debug when client and server have a different understanding of what the count is. Considering that, please confirm you'd rather it be omitted and I'll update.

tfpauly: I'd love to avoid having the sequence number as an extra field, just to reduce the number of things we're adding to the wire format. I get that there's a bit more state to keep, but this is a reliable stream.

draft-ietf-masque-quic-proxy.md
ehaydenr: If the client just sent a fin on the client-to-proxy HTTP stream, that wouldn't give the proxy indication that something bad happened which is unfortunate. If we went with a reset, what's the most appropriate? H3_GENERAL_PROTOCOL_ERROR? 🤔

draft-ietf-masque-quic-proxy.md
DavidSchinazi: Yeah but that's what we've done with other capsules, here's from [connect-ip](https://www.rfc-editor.org/rfc/rfc9484#section-4.7.2-8):

If any of the capsule fields are malformed upon reception, the receiver of the capsule MUST follow the error-handling procedure defined in [Section 3.3](https://rfc-editor.org/rfc/rfc9297#section-3.3) of [[HTTP-DGRAM](https://www.rfc-editor.org/rfc/rfc9484#RFC9297)].

draft-ietf-masque-quic-proxy.md
DavidSchinazi: My gut feeling is that as an implementer I would keep track of the last sequence number anyway, and manually validate that the peer is sending them in correct order. So removing the field means less code

draft-ietf-masque-quic-proxy.md
DavidSchinazi: Is that the guidance? Or should it be that the client just can't migrate any more? Happy to move this to a separate issue if you prefer

ehaydenr: Migration is only relevant when you're doing forwarding mode. Limiting CID registration is still useful when there's no forwarding because the proxy may want to limit the number of CIDs considered for port reuse

DavidSchinazi: Oh right I had it backwards. If the client runs out of CIDs in this way, it's the CIDs on the target-to-proxy leg right? So the consequence is that the client can't send new connection IDs to the target, but the connection could keep working?

DavidSchinazi: I see. OK then I think it's worth it to explain this scenario and tell the client that they will need to abort the connection if after some timeout they're still blocked in this state

DavidSchinazi (approved):
Thanks!

draft-ietf-masque-quic-proxy.md
```diff
@@ -468,7 +476,10 @@ using HTTP Datagrams with the context ID set to zero as defined in
 The REGISTER_CLIENT_CID ({{capsule-reg-client}}) and REGISTER_TARGET_CID
 ({{capsule-reg-target}}) capsule types allow a client to inform
 the proxy about a new client CID or a new target CID,
-respectively. These capsule types MUST only be sent by a client.
+respectively. These capsule types MUST only be sent by a client. These capsule
+types share a sequence number space which allows the proxy to limit the
+number of active registrations. The first registration has sequence number 0
+and subsequent registrations increment the sequence number by 1.
```
tfpauly, line 482:
```suggestion
number of active registrations. The first registration (of either client CID or target CID)
has sequence number 0, and subsequent registrations increment the sequence number
by 1.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -753,6 +790,19 @@ when the HTTP stream's network path changes to avoid linkability across network
...
 receiving an HTTP response from the proxy.
 
+Connection ID registrations are subject to a proxy-advertised limit. Each registration
+has a corresponding sequence number. The client MUST NOT send a registration
+capsule with a sequence number greater than what the proxy advertises via the
+MAX_CONNECTION_IDS capsule. The initial MAX_CONNECTION_IDS value is 1, allowing both
+sequence numbers 0 and 1 for a total of two registrations without receiving a
+MAX_CONNECTION_IDS capsule from the proxy.
+
+Clients which cannot register new connection IDs within a reasonable timeout due to
```
tfpauly, line 800:
```suggestion
Clients that cannot register new connection IDs within a reasonable time due to
```

tfpauly (approved):

PR #119, opened 2025-02-28
guidance for disabling forwarding mode

### LINKED ISSUE #91: No way of disabling forwarding mode half way through stream
Connection ID registration capsules allow the client to communicate client and target connection IDs to the proxy. This enables both port reuse on the proxy and the option to use forwarding mode.

When the head `Proxy-QUIC-Forwarding=?1` was mutually communicated, the client **may** send in forwarding mode, but **must** be able to receive in forwarding mode. As a client, if "mid stream" I decide a don't want to do forwarding anymore, I don't have a way to communicate this to the proxy. CloseClientCid/CloseTargetCid capsules don't work here because that would break the proxy's ability to distinguish QUIC packets on a reused UDP port.

ehaydenr:
closes #91

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -804,7 +804,7 @@ stream with error code NO_ERROR. This may happen, for example, if the target ser
 sends a NEW_CONNECTION_ID frame with Sequence Number and Retire Prior To equal to the
 same value.
 
-Clients may cease receiving with forwarded mode over an existing tunnel while
+Clients can cease receiving with forwarded mode over an existing tunnel while
 retaining the same client-to-target connection by creating a new tunnel with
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -804,6 +804,10 @@ stream with error code NO_ERROR. This may happen, for example, if the target ser
 sends a NEW_CONNECTION_ID frame with Sequence Number and Retire Prior To equal to the
 same value.
 
+Clients may cease receiving with forwarded mode over an existing tunnel while
```
tfpauly, line 807:
```suggestion
Clients can cease receiving with forwarded mode over an existing tunnel while
```

tfpauly (approved):

DavidSchinazi (commented):
LGTM

Issue #123, opened 2025-03-21
Proxy-QUIC-Forwarding stuff


LPardue:
[Draft 05](https://www.ietf.org/archive/id/draft-ietf-masque-quic-proxy-05.html#section-3) says some stuff that I'll comment in line on:

> "Proxy-QUIC-Forwarding" is an Item Structured Header [[RFC8941](https://www.ietf.org/archive/id/draft-ietf-masque-quic-proxy-05.html#RFC8941)]. Its value MUST be a Boolean.
> If the client wants to enable QUIC packet forwarding for this request, it sets the value to "?1". If it doesn't want to enable forwarding, but instead only provide information about QUIC Connection IDs for the purpose of allowing the proxy to share a proxy-to-target 4-tuple, it sets the value to "?0".

Fine

> The client MUST add an "accept-transform" parameter whose value is an sf-string containing the supported packet transforms ([Section 6.3](https://www.ietf.org/archive/id/draft-ietf-masque-quic-proxy-05.html#transforms)) in order of descending preference, separated by commas. If the proxy receives a "Proxy-QUIC-Forwarding" header without the "accept-transform" parameters, it MUST ignore the header and respond as if the client had not sent the "Proxy-QUIC-Forwarding" header.

nit: `sf-string` is the old terminology we don't like please use `String`.

major: IIUC the value is a comma-separated list of values. The only currently defined values are `identity` and `scramble-dt` (to be changed to `scramble` later). So a couple of point here:

1. What does it mean to receive a value that is not known? For example, what if I send `poached`? I presume its not an issue and should be ignored but nothing lept out at me
2. How are you going to deal with whitespace in these values? E.g. if I send `"scramble, identity"` or `"   scramble, identity    "`
3. How do you deal with empty entries such as `"scramble, , identity"`?

> If the proxy supports QUIC-aware proxying, it will include the "Proxy-QUIC-Forwarding" header in successful HTTP responses. The value indicates whether or not the proxy supports forwarding. If the client does not receive this header in responses, the client SHALL assume that the proxy does not support this extension.
> The proxy MUST include a "transform" parameter whose value is an sf-string indicating the selected transform. If the proxy does not recognize or accept any of the transforms offered by the client, it MUST omit this parameter and set the header field value to "?0", or omit the header entirely.

major: 

4. does the value have to be identical to one of the values that client sent? It seems implicit it should - what does a client do if the server is b0rked (i.e. should it abort the request?)

[...]

PR #126, opened 2025-06-12
make receiving the wrong parameter undefined

ehaydenr:
addresses feedback from #123


ehaydenr:
@DavidSchinazi , would it be acceptable to say it's not defined by this document?

DavidSchinazi:
That's the same thing as undefined :)

The issue is that if you don't define it, then some implementations might end up sending it to mean something and others to mean something else. And then various implementations will react differently when receiving it.

Isn't it just simpler to specify it?

ehaydenr:
@DavidSchinazi , I've updated the guidance in #128. Please take a look there.

Issue #115, opened 2024-09-20
Active attack on IV for SCRAMBLE transform


gloinul:
An attacker that has access and capability to modify packets part of an end-to-end QUIC connection prior to QUIC aware forwarding with SCRAMBLE transform can enable defeating the scramble transform. This is done by manipulating the part of the QUIC packet that will be used as IV. The attack modify some packets prior to reaching either the MASQUE client or the MASQUE proxy so this set of packets all have the same bit-value in the parts that will be used as IV. This will result in the E2E packet will be dropped at arrival at the endpoint. 

By setting the same IV two effects will be achieved:
1) The scrambled IV after AES-ECB application will have the same value in all the packets, enabling an attacker that captures all packets after the Masque tunnel ingress's forwarding and transformation to look for the packet with the same but unknown bit-pattern in a given location within the packet
2) Due to the use of the same IV in AES-CTR mode between the multiple packet that actual scrambling can be defeated. Although I think this in most cases would have limited value as it would only allow recovering the End-to-End packet that this attacker already have access to. I think 1) is sufficient to enable the important linking of input and output 5-tuples and CIDs. 

The weakness here is that the attacker can chose the IV, which is not true in the QUIC header protection application of using part of the packet as IV. But, in this case the MASQUE proxy or client can't verify the end-to-end forwarded packets integrity before choosing to use it as IV. 

I don't know if this attack is serious enough for scramble to change its solution, but at a minimal the potential attack and the weakness it creates needs to be documented. But likely a bit of thought should be put into if another solution can be found to avoid this attack.

gloinul:
I think the merged text looks good and is sufficient warning for the issue. So lets close this issue.

PR #129, opened 2025-06-27
mention active attack on scramble IV

ehaydenr:
Discussed #115 with @tfpauly offline. The follow text is his suggestion (which I agree with).


DavidSchinazi:
LGTM

PR #131, opened 2025-07-07
Add section explaining server preferred addresses

### LINKED ISSUE #86: Recommendations for Address Migration
The best practice when clients migrate is not immediately obvious.

As the proxy's source IP does not change, changing the connection ID is not strictly necessary between proxy and target.

If a client migrates, clearly the virtual connection ID should change, to avoid linkability along the client/proxy path. But it's not totally clear to me if the "real" connection IDs should change.

If the connection IDs do not change, and an attacker is otherwise able to correlate client-proxy flows with proxy-target flows, then that makes things linkable.

However, changing the connection ID means short header packets start with a new connection ID, which is the signature of a migrating connection. This could actually allow an observer to correlate a client-proxy flow with a proxy-target flow.

I don't have a solution but I'm happy to help think through the tradeoffs here.

### LINKED ISSUE #113: How to handle preferred address migration?
I am wondering what is supposed to happen if the server sends a "preferred address" parameter. My understanding of the draft is that all packets produced by the client are sent to the IP address and port specified in the CONNECT-UDP request. It follows that if the client wants to heed the "preferred address" parameter, it should set up a new proxy connection for that. Is that a correct understanding? Should we explain that in the draft?

tfpauly:
Closes #113
Closes #86

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -1227,7 +1227,9 @@ Note that clients can be aware of the target address being used for the original
 proxy request by looking at `next-hop` parameter on a `Proxy-Status` header
 sent by the proxy in its response ({{Section 2.1.2 of ?RFC9209}}). This allows
 a client to know if it ought to migrate to the preferred address, or is already
-connected to the preferred address.
+connected to the preferred address. To support the ability of clients to do this, proxies
+implementing this specification SHOULD send the `Proxy-Status` header in responses
+and include the `next-hop` parameter.
 
 ## Handling ECN Marking {#ecn}
 
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -1003,7 +1003,8 @@ into VCIDs, and the contents of packets are either left the same, or modified
 
 Forwarded mode also raises special considerations for handling connection
 maintenance ({{maintenance}}), connection migration ({{migration}}),
-ECN markings ({{ecn}}), and stateless resets ({{resets}}).
+server preferred addresses ({{preferred-address}}), ECN markings ({{ecn}}),
+and stateless resets ({{resets}}).
 
 ## Sending With Forwarded Mode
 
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -1209,6 +1210,25 @@ virtual connection IDs to prevent packets with the same connection ID bytes bein
...
+## Handling Server Preferred Addresses {#preferred-address}
+
+QUIC allows servers to tell clients about a preferred address the server
+would like to use. When this happens, the client can migrate to the preferred
+address.
+
+When a client using a proxy wants to migrate to the preferred address of the
+target server, it needs to create a new CONNECT request to the proxy (using
+the method defined in {{CONNECT-UDP}}) targeting the preferred address of the
+target. This is the behavior clients using a proxy will have regardless of using
```
LPardue, line 1222: I'm a bit out of date but how does it indicate the preferred address, by passing an IP rather than a name?

draft-ietf-masque-quic-proxy.md
```diff
@@ -1209,6 +1210,25 @@ virtual connection IDs to prevent packets with the same connection ID bytes bein
 used over multiple network paths. Note that the client CID and target CID
 can stay the same while the target VCID and client VCID change.
 
+## Handling Server Preferred Addresses {#preferred-address}
+
+QUIC allows servers to tell clients about a preferred address the server
+would like to use. When this happens, the client can migrate to the preferred
```
DavidSchinazi, line 1216: nit: add reference to Section 9.6 of RFC 9000

draft-ietf-masque-quic-proxy.md
```diff
@@ -1209,6 +1210,25 @@ virtual connection IDs to prevent packets with the same connection ID bytes bein
...
+address.
+
+When a client using a proxy wants to migrate to the preferred address of the
+target server, it needs to create a new CONNECT request to the proxy (using
```
DavidSchinazi, line 1220: nit: this would match the title of 9298 better

```suggestion
target server, it needs to create a new UDP proxying request to the proxy (using
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1209,6 +1210,25 @@ virtual connection IDs to prevent packets with the same connection ID bytes bein
...
+the method defined in {{CONNECT-UDP}}) and using the preferred IP address of the
+target as the host to which to connect. This is the behavior clients using a proxy will have regardless of using
+the QUIC-aware mechanisms defined in this document. From the proxy's perspective,
+the migrating request is separate from the original request.
+
+Note that clients can be aware of the target address being used for the original
+proxy request by looking at `next-hop` parameter on a `Proxy-Status` header
+sent by the proxy in its response ({{Section 2.1.2 of ?RFC9209}}). This allows
+a client to know if it ought to migrate to the preferred address, or is already
+connected to the preferred address.
```
DavidSchinazi, line 1230: Maybe it would help to add something here like "For this reason, proxies implementing this specification SHOULD send the `Proxy-Status` header and include the `next-hop` parameter.

tfpauly:
```suggestion
connected to the preferred address. To support the ability of clients to do this, proxies
implementing this specification SHOULD send the `Proxy-Status` header in responses
and include the `next-hop` parameter.
```

Issue #79, opened 2023-03-29
Forwarded packets are not part of the MASQUE connection


martinduke:
This is editorial, but I would like the "forwarding mode" section to unambiguously state that the forwarding-mode packets are not part of the outer connection for any purpose (congestion control, flow control, acks, etc). This is implied by the keepalive section, but ought to be explicit.

martinduke:
Priority is another thing where one might be tempted to apply connection rules.

tfpauly:
Closed by #138

PR #138, opened 2025-07-18
unambiguously state forwarded mode packets are independent

ehaydenr:
addresses #79


tfpauly (approved):

martinduke:
I would have liked a few examples of things in the outer connection it does not affect, as discussed in the issue, but I can live with this.

PR #166, opened 2026-02-26
Update MAX_CONNECTION_IDS capsule description

### LINKED ISSUE #157: MAX_CONNECTION_IDS advertises an inclusive sequence-number maximum
For readers coming from QUIC v1, there’s a small semantic mismatch: QUIC’s MAX_* limits are typically understood as exclusive with respect to an index/offset (e.g., MAX_DATA=4 ⇒ offsets < 4, i.e., 4 bytes total), whereas MAX_CONNECTION_IDS communicates an inclusive maximum over a registration sequence number (e.g., initial maximum = 1 permits registrations with sequence numbers 0 and 1).

I don’t have a strong preference, but this can surprise implementers.

Unless there is a precedence to use inclusive maximums inside the masque WG, it might make sense to use exclusive maximums to minimize surprise.

### LINKED ISSUE #162: Can a proxy retract CID registration space by advertising a smaller maximum with MAX_CONNECTION_IDS?
The draft does not state whether a proxy may send a MAX_CONNECTION_IDS capsule carrying a maximum smaller than previously advertised, which creates confusion about whether proxies can retract registration space once advertised.

It might make sense to explicitly prohibit sending smaller values.

tfpauly:
Closes #157 
Closes #162

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -815,9 +815,9 @@ Maximum Connection IDs
 is allowed to request. For example, if the value is 4, the maximum allowed sequence
 number for a connection ID registration would be 3.
 
-The value advertised in the capsule MUST NOT be smaller than any value previously
-sent in a MAX_CONNECTION_IDS capsule, and MUST not be smaller than the initial
-allowed limit of 2. Thus, the value sent in the capsule MUST NOT be less than 3.
+The value advertised in the capsule MUST be greater than any value previously
+sent in a MAX_CONNECTION_IDS capsule, and MUST be greater than the initial
+allowed limit of 2. Thus, any value sent in the capsule MUST at least 3.
 
 Clients receiving a MAX_CONNECTION_IDS capsule with an invalid value
 MUST reset the stream with H3_DATAGRAM_ERROR error code.
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -552,12 +552,13 @@ either rejecting the proxy-chosen client VCID or no longer
 needs the connection ID registered.
 
 The MAX_CONNECTION_IDS capsule type {{capsule-max-cids}} MUST only be sent by the
-proxy. It indicates to the client the maximum permitted sequence number for
+proxy. It indicates to the client the maximum number of permitted sequence numbers for
 connection ID registrations. This allows the proxy to limit the number of active
-registrations. The initial maximum is 1, allowing the client to send 2 registrations,
-one with sequence number 0 and another with sequence number 1. A proxy MUST NOT
-send a MAX_CONNECTION_IDS capsule with a value less than 1. Clients receiving a
-MAX_CONNECTION_IDS capsule with a value less than 1 MUST reset the stream with
+registrations. The initial maximum is 2, allowing the client to send 2 registrations,
+one with sequence number 0 and another with sequence number 1. MAX_CONNECTION_IDS are
+only sent to increase the limit, so since the initial limit is 2, a proxy MUST NOT
+send a MAX_CONNECTION_IDS capsule with a value less than 3. Clients receiving a
+MAX_CONNECTION_IDS capsule with a value less than 3 MUST reset the stream with
 H3_DATAGRAM_ERROR error code.
 
 When port sharing {{port-sharing-header}} is supported, the client MUST register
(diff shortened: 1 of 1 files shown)


tfpauly:
@kazuho please check too!

draft-ietf-masque-quic-proxy.md
```diff
@@ -552,12 +552,13 @@ either rejecting the proxy-chosen client VCID or no longer
 needs the connection ID registered.
 
 The MAX_CONNECTION_IDS capsule type {{capsule-max-cids}} MUST only be sent by the
-proxy. It indicates to the client the maximum permitted sequence number for
+proxy. It indicates to the client the maximum number of permitted sequence numbers for
 connection ID registrations. This allows the proxy to limit the number of active
```
```suggestion
proxy. It indicates to the client the cumulative number of connection ID registrations the client is allowed to request. This allows the proxy to limit the number of active
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -797,21 +798,30 @@ to 20 bytes, but QUIC invariants allow up to 255 bytes.
...
+  Maximum Connection IDs (i)
 }
 ~~~
 {: #fig-capsule-max-connection-ids title="Maximum Connection IDs Capsule Format"}
 
-Maximum Sequence Number
-: The maximum permitted sequence number for connection ID registrations. This MUST
-NOT be less than 1.
+Maximum Connection IDs
+: A count of the cumulative number of connection ID registrations the the client
```
```suggestion
: A count of the cumulative number of connection ID registrations the client
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -552,12 +552,12 @@ either rejecting the proxy-chosen client VCID or no longer
...
-proxy. It indicates to the client the maximum permitted sequence number for
-connection ID registrations. This allows the proxy to limit the number of active
-registrations. The initial maximum is 1, allowing the client to send 2 registrations,
-one with sequence number 0 and another with sequence number 1. A proxy MUST NOT
-send a MAX_CONNECTION_IDS capsule with a value less than 1. Clients receiving a
-MAX_CONNECTION_IDS capsule with a value less than 1 MUST reset the stream with
+proxy. It indicates to the client the cumulative number of connection ID registrations the client is allowed to request. This allows the proxy to limit the number of active
+registrations. The initial maximum is 2, allowing the client to send 2 registrations,
+one with sequence number 0 and another with sequence number 1. MAX_CONNECTION_IDS are
+only sent to increase the limit, so since the initial limit is 2, a proxy MUST NOT
```
kazuho:
```suggestion
only sent to increase the limit. Since the initial limit is 2, a proxy MUST NOT
```
Editorial: It might be easier to read if we split the sentence.

draft-ietf-masque-quic-proxy.md
```diff
@@ -797,21 +797,30 @@ to 20 bytes, but QUIC invariants allow up to 255 bytes.
...
-: The maximum permitted sequence number for connection ID registrations. This MUST
-NOT be less than 1.
+Maximum Connection IDs
+: A count of the cumulative number of connection ID registrations the client
+is allowed to request. For example, if the value is 4, the maximum allowed sequence
+number for a connection ID registration would be 3.
+
+The value advertised in the capsule MUST NOT be smaller than any value previously
+sent in a MAX_CONNECTION_IDS capsule, and MUST not be smaller than the initial
+allowed limit of 2. Thus, the value sent in the capsule MUST NOT be less than 3.
```
kazuho: I think we have a contradiction here. The examples here and above prohibit sending MAX_CONNECTION_IDS capsule containing the same value. But the MUST NOT only prohibits sending a smaller value, while allowing the same value to be advertised twice.

Considering that the capsule is sent on a reliable stream, my two cents go to "MUST increase." The rationale for not using "MUST NOT decrease" in consistency with the frames defined in RFC 9000 would be that the validation rules have to be different anyways and thus we can be even more strict - for frames that might get dropped, it is not practical to reject a particular behavior, whereas for capsules that is possible due to them sent over a reliable stream.

kazuho (commented):
Thank you for the update! Looks mostly good.

draft-ietf-masque-quic-proxy.md
tfpauly: Yes, I totally agree! That was my intent, but I just wasn't clear in my text.

tfpauly: Updated, please check

kazuho (approved):
Thank you! Looks perfect to me.

ehaydenr (approved):

DavidSchinazi (approved):

draft-ietf-masque-quic-proxy.md
```diff
@@ -797,21 +797,30 @@ to 20 bytes, but QUIC invariants allow up to 255 bytes.
...
-: The maximum permitted sequence number for connection ID registrations. This MUST
-NOT be less than 1.
+Maximum Connection IDs
+: A count of the cumulative number of connection ID registrations the client
+is allowed to request. For example, if the value is 4, the maximum allowed sequence
+number for a connection ID registration would be 3.
+
+The value advertised in the capsule MUST be greater than any value previously
+sent in a MAX_CONNECTION_IDS capsule, and MUST be greater than the initial
+allowed limit of 2. Thus, any value sent in the capsule MUST at least 3.
```
```suggestion
allowed limit of 2. Thus, any value sent in the capsule MUST be at least 3.
```

PR #169, opened 2026-03-02
Add reason codes to Close CID capsules and clarify semantics

### LINKED ISSUE #160: Short client CIDs can induce loops
[Section 5.8](https://ietf-wg-masque.github.io/draft-ietf-masque-quic-proxy/draft-ietf-masque-quic-proxy.html#section-5.8) states that “a proxy MAY choose to reject all registrations for very short Connection IDs as conflicts, in anticipation of future conflicts.” The draft also says in [section 5.10](https://ietf-wg-masque.github.io/draft-ietf-masque-quic-proxy/draft-ietf-masque-quic-proxy.html#section-5.10-10) that, in order to permit the client to change client-to-target connection IDs, the proxy SHOULD send MAX_CONNECTION_IDS capsules allowing additional registrations.

An impatient proxy implementation might combine these two clauses and unconditionally reject client CIDs shorter than some threshold (e.g., <9 bytes) while immediately sending MAX_CONNECTION_IDS, effectively inviting the client to retry. Likewise, a client might keep retrying with a merely “long enough” CID (e.g., 8 bytes) before giving up. The net effect is a loop.

It could make sense to clarify that proxies SHOULD NOT send a MAX_CONNECTION_IDS capsule solely in response to rejecting a CID registration (e.g., via CLOSE_CLIENT_CID) for being too short. This avoids creating retry loops when the proxy’s policy intends a permanent rejection.

### LINKED ISSUE #165: Clarify MAX_CONNECTION_IDS interaction with re-registering a CID
Clarify that if the same CID is registered (REGISTER_CLIENT_CID), it doesn't increment any sequence number, so doesn't go over MAX_CONNECTION_IDS.

### LINKED ISSUE #167: Bump capsule numbers
We've updated the behavior of MAX_CONNECTION_IDs, we need new capsule types

### LINKED ISSUE #168: REGISTER_CLIENT_CID doesn't have Stateless Reset token, but Example does
No Stateless Reset Token here: https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/blob/main/draft-ietf-masque-quic-proxy.md#register_client_cid-capsule-reg-client

Present in example exchange: https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/blob/main/draft-ietf-masque-quic-proxy.md#example-exchange

ehaydenr:
- Add Reason field to CLOSE_CLIENT_CID and CLOSE_TARGET_CID capsule format with new IANA registry defining RETIRED, TOO_SHORT, and CONFLICT reason codes
- Clarify that every registration consumes a sequence number, including rejected registrations and re-registrations
- Distinguish client vs proxy behavior: clients send CLOSE to retire CIDs, proxies send CLOSE only to reject registrations
- Add constraint that proxies MUST NOT close already-acknowledged CIDs
- Update capsule type values from 0xffe6xx to 0xffe7xx
- Fix example: move Stateless Reset Token to REGISTER_TARGET_CID

Closes #168, #167, and #160

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -897,7 +897,7 @@ receiving an HTTP response from the proxy.
 If the client receives a VCID it cannot use, it may re-register the same
 client CID with a reason code indicating why the previous VCID was unusable.
 The proxy SHOULD use this information to select a more suitable VCID. If the
-reason is TOO_SHORT, the proxy SHOULD select a longer VCID. If the reason is
+reason is TOO_SHORT, the proxy MUST either select a longer VCID or close the registration. If the reason is
 CONFLICT, the proxy MUST select a different VCID.
 
 Connection ID registrations are subject to a proxy-advertised limit. Each registration
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -520,9 +520,10 @@ The REGISTER_CLIENT_CID ({{capsule-reg-client}}) and REGISTER_TARGET_CID
 the proxy about a new client CID or a new target CID,
 respectively. These capsule types MUST only be sent by a client. These capsule
 types share a sequence number space which allows the proxy to limit the
-number of active registrations. The first registration (of either client CID or target CID)
+number of registrations. The first registration (of either client CID or target CID)
 has sequence number 0, and subsequent registrations increment the sequence number
-by 1.
+by 1. Every registration capsule consumes a sequence number, including registrations
+that are rejected and re-registrations of a previously registered connection ID.
 
 The ACK_CLIENT_CID ({{capsule-ack-client}}) and ACK_TARGET_CID
 ({{capsule-ack-target}}) capsule types are sent by the proxy to the client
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-quic-proxy.md
```diff
@@ -1644,6 +1685,22 @@ Specification Required policy (Section 4.6 of [IANA-POLICY]).
...
+This document establishes a new registry, "CID Capsule Reason Codes",
+for reason codes used in REGISTER_CLIENT_CID, REGISTER_TARGET_CID,
+CLOSE_CLIENT_CID, and CLOSE_TARGET_CID capsules,
+in <[](https://www.iana.org/assignments/masque/masque.xhtml)>.
+Registrations in this registry are assigned using the
+Specification Required policy (Section 4.6 of [IANA-POLICY]).
+
+| Value | Name       | Description                                      | Specification |
+|:------|:-----------|:-------------------------------------------------|:--------------|
+| 0x00  | NONE       | Normal operation                                 | This Document |
```
```suggestion
| 0x00  | NO_ERROR       | No error, normal operation                                 | This Document |
```

tfpauly: Or DEFAULT

draft-ietf-masque-quic-proxy.md
```diff
@@ -621,6 +637,10 @@ Register Target CID Capsule {
 ~~~
 {: #fig-capsule-register-target-cid title="Register Target CID Capsule Format"}
 
+Reason:
+: The reason for this registration. This MUST be NONE (0x00).
```
```suggestion
: The reason for this registration. This MUST be NONE (0x00), although a different value can be used for future extensions.
```

draft-ietf-masque-quic-proxy.md
ehaydenr: updated

draft-ietf-masque-quic-proxy.md
```diff
@@ -621,6 +637,10 @@ Register Target CID Capsule {
 ~~~
 {: #fig-capsule-register-target-cid title="Register Target CID Capsule Format"}
 
+Reason:
+: The reason for this registration. This MUST be DEFAULT (0x00).
```
```suggestion
: The reason for this registration. This MUST be DEFAULT (0x00), although a different value can be specified by future extensions.
```

tfpauly (approved):

draft-ietf-masque-quic-proxy.md
ehaydenr: updated

draft-ietf-masque-quic-proxy.md
ehaydenr: updated

draft-ietf-masque-quic-proxy.md
```diff
@@ -866,6 +894,12 @@ when the HTTP stream's network path changes to avoid linkability across network
 paths. Note that initial REGISTER_CLIENT_CID capsules MAY be sent prior to
 receiving an HTTP response from the proxy.
 
+If the client receives a VCID it cannot use, it may re-register the same
+client CID with a reason code indicating why the previous VCID was unusable.
+The proxy SHOULD use this information to select a more suitable VCID. If the
+reason is TOO_SHORT, the proxy SHOULD select a longer VCID. If the reason is
```
DavidSchinazi: Why is this a SHOULD?

draft-ietf-masque-quic-proxy.md
```diff
@@ -1644,6 +1687,22 @@ Specification Required policy (Section 4.6 of [IANA-POLICY]).
...
 {: #iana-packet-transforms-table title="Initial Packet Transform Names"}
 
+## CID Capsule Reason Codes {#iana-cid-reasons}
+
+This document establishes a new registry, "CID Capsule Reason Codes",
+for reason codes used in REGISTER_CLIENT_CID, REGISTER_TARGET_CID,
+CLOSE_CLIENT_CID, and CLOSE_TARGET_CID capsules,
+in <[](https://www.iana.org/assignments/masque/masque.xhtml)>.
+Registrations in this registry are assigned using the
+Specification Required policy (Section 4.6 of [IANA-POLICY]).
```
DavidSchinazi:
```suggestion
This registry governs a 62-bit space and operates under the QUIC
registration policy documented in {{Section 22.1 of QUIC}}. This new registry
includes the common set of fields listed in {{Section 22.1.1 of QUIC}}. In
addition to those common fields, all registrations in this registry MUST include
a "Name" field that contains a short name or label for the Reason.

Permanent registrations in this registry are assigned using the Specification
Required policy ({{Section 4.6 of !IANA-POLICY=RFC8126}}), except for values
between 0x00 and 0x3f (in hexadecimal; inclusive), which are assigned using
Standards Action or IESG Approval as defined in {{Sections 4.9 and 4.10 of
IANA-POLICY}}.
```

DavidSchinazi (approved):

draft-ietf-masque-quic-proxy.md
tfpauly: How about
```suggestion
reason is TOO_SHORT, the proxy MUST either select a longer VCID or close the registration. If the reason is
```

PR #170, opened 2026-06-29
migration clarifications

### LINKED ISSUE #158: Split CID registration space
IIUC, client CIDs and target CIDs share the same registration space controlled by MAX_CONNECTION_IDS. However, that might not be ideal, especially when the space is tight.

Typically, a proxy wants to retain two client CIDs: one used actively for forwarding packets and another kept as a backup for when the client migrates to a new path. Once the client migrates, the proxy would retire the old client CID and send a MAX_CONNECTION_IDS capsule, asking the client to register a new client CID.

However, the intent behind MAX_CONNECTION_IDS is unclear: the client might instead consume that space by registering a new target CID.

One could argue the client should infer proxy intent from how many CIDs are currently registered for each role, but that’s fragile and adds complexity. Maybe it makes sense to split the space and define two limits, e.g., MAX_CLIENT_CONNECTION_IDS and MAX_TARGET_CONNECTION_IDS.

### LINKED ISSUE #159: Provide advice on how many CIDs a proxy should retain
The draft is currently light on guidance for how many CIDs a proxy ought to accept and retain; the only implicit advice is the initial maximum of 2 (i.e., two registrations across the shared sequence space). Some operational guidance on a desirable minimum would help align implementations.

I might recommend that a proxy retain at least 2 client CIDs (active + standby for migration; see #158) and 1 target CID. Put differently: if the client/target registration space remains shared, advise a minimum MAX_CONNECTION_IDS that allows at least 3 registrations rather than 2, and adjust the default accordingly.

ehaydenr:


draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
 initiating active migration, clients are no longer able to send forwarded mode
 packets since the proxy will have removed forwarding rules. Clients can proceed with
 tunnelled mode or can request new forwarding rules via REGISTER_CLIENT_CID and
-REGISTER_TARGET_CID capsules. Each of the acknowledging capsules will contain new
-virtual connection IDs to prevent packets with the same connection ID bytes being
-used over multiple network paths. Note that the client CID and target CID
-can stay the same while the target VCID and client VCID change.
+REGISTER_TARGET_CID capsules. Requesting new forwarding rules like this for the same
+client-to-target CIDs, is a re-registration and does not increment the sequence
+number or have any impact on MAX_CONNECTION_IDs. Each of the acknowledging capsules
```
```suggestion
REGISTER_TARGET_CID capsules. Requesting new forwarding rules in this way (registering the same
client-to-target CIDs again) is a re-registration and does not increment the sequence
number or have any impact on MAX_CONNECTION_IDs. Each of the acknowledging capsules
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+will contain new virtual connection IDs to prevent packets with the same connection
+ID bytes being used over multiple network paths. Note that the client CID and target
+CID can stay the same while the target VCID and client VCID change. Importantly,
+the client and proxy do not send a CLOSE_CLIENT_CID capsule because that would also
+remove the registration of the CID for the purpose of port sharing - breaking the
```
```suggestion
remove the registration of the CID for the purpose of port sharing, potentially breaking the
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+tunnelled path.
+
+### Passive Migration Steps {#passive-migration-steps}
+
+1. Client registers connection IDs via REGISTER_CLIENT_CID/REGISTER_TARGET_CID
+1. Client experiences NAT rebinding
+1. Proxy recognfigures forwarding rules to reuse the same virtual CID on the new network path.
```
```suggestion
1. Proxy reconfigures forwarding rules to reuse the same virtual CID on the new network path.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+Forwarding rules activated upon validation of the client-proxy network path.
+
+No capsules were exchanged due to passive migration.
```
```suggestion
No capsules are exchanged in response to passive migration.
```

tfpauly (approved):

ehaydenr:
@kazuho, does this change address your feedback?

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
 initiating active migration, clients are no longer able to send forwarded mode
 packets since the proxy will have removed forwarding rules. Clients can proceed with
 tunnelled mode or can request new forwarding rules via REGISTER_CLIENT_CID and
-REGISTER_TARGET_CID capsules. Each of the acknowledging capsules will contain new
-virtual connection IDs to prevent packets with the same connection ID bytes being
-used over multiple network paths. Note that the client CID and target CID
-can stay the same while the target VCID and client VCID change.
+REGISTER_TARGET_CID capsules. Requesting new forwarding rules in this way (registering the same
+client-to-target CIDs again) is a re-registration and does not increment the sequence
+number or have any impact on MAX_CONNECTION_IDs. Each of the acknowledging capsules
```
kazuho: I wonder if this is contradictory to the existing statements elsewhere:
* Every registration capsule consumes a sequence number, including registrations that are rejected and re-registrations of a previously registered connection ID. ([Section 5](https://ietf-wg-masque.github.io/draft-ietf-masque-quic-proxy/draft-ietf-masque-quic-proxy.html#section-5-3))
* Note that re-registrations consume sequence numbers like any other registration. ([Section 5.9](https://ietf-wg-masque.github.io/draft-ietf-masque-quic-proxy/draft-ietf-masque-quic-proxy.html#name-client-considerations)).

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
 packets with the same connection ID bytes over multiple network paths. After
 initiating active migration, clients are no longer able to send forwarded mode
 packets since the proxy will have removed forwarding rules. Clients can proceed with
 tunnelled mode or can request new forwarding rules via REGISTER_CLIENT_CID and
-REGISTER_TARGET_CID capsules. Each of the acknowledging capsules will contain new
-virtual connection IDs to prevent packets with the same connection ID bytes being
-used over multiple network paths. Note that the client CID and target CID
-can stay the same while the target VCID and client VCID change.
+REGISTER_TARGET_CID capsules. Requesting new forwarding rules in this way (registering the same
+client-to-target CIDs again) is a re-registration and does not increment the sequence
```
kazuho: Previous sentence talks about REGISTER_CLIENT_CID and REGISTER_TARGET_CID, but this sentence is about client-to-target CIDs carried specifiallly by REGISTER_TARGET_CID. Therefore, I think it would make sense to clarify the subject.

```suggestion
REGISTER_TARGET_CID capsules. The REGISTER_TARGET_CID capsules might contain the
same client-to-target CID as before; this is a re-registration and does not increment the sequence
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
 
 When operating in forwarded mode, the proxy reconfigures or removes forwarding
 rules as the network path between the client and proxy changes. In the event of
```
kazuho: As the paragraph is long, I wonder it makes sense to split them, having one paragraph for passive and active migration.

Separately, this is nothing more than a matter of taste, but it might make sense to remind the readers what "fowarding rules" mean.

```suggestion
rules (i.e., the mappings for VCIDs) as the network path between the client and proxy changes.

In the event of
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+will contain new virtual connection IDs to prevent packets with the same connection
+ID bytes being used over multiple network paths. Note that the client CID and target
+CID can stay the same while the target VCID and client VCID change. Importantly,
+the client and proxy do not send a CLOSE_CLIENT_CID capsule because that would also
+remove the registration of the CID for the purpose of port sharing, potentially breaking the
+tunnelled path.
+
+### Passive Migration Steps {#passive-migration-steps}
+
+1. Client registers connection IDs via REGISTER_CLIENT_CID/REGISTER_TARGET_CID
```
kazuho:
```suggestion
1. Client and proxy exchange connection IDs via REGISTER_CLIENT_CID/REGISTER_TARGET_CID/ACK_CLIENT_CID/ACK_TARGET_CID and start using the forwarded mode.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+
+### Active Migration Steps {#active-migration-steps}
+
+1. Client registers client-target connection IDs while on client-proxy network path "Wi-Fi"
+1. Client probes client-proxy network path "Cellular" and migrates
+1. Proxy removes forwarding rules on path "Wi-Fi". The forwarding rules are removed, but
+the registration remains active in order to continue to support port sharing.
+1. Client re-registers the original client-target connection IDs while on client-proxy network path
+"Cellular". This solicits new virtual CIDs from the proxy and does not count as an additional
+sequence number towards MAX_CONNECTION_IDs.
```
kazuho: Having these steps written down helps a lot! Thank you for doing this.

Since QUIC version 1 does not talk specifically about having just 2 paths ([it allows more](https://datatracker.ietf.org/doc/html/rfc9000#section-8.2-8)), I wonder if it makes sense to generalize. Also, it could make sense to clarify how a proxy might detect active migration (RFC 9000 does not describe how, as QUIC version 1 servers do not need to distnguish between active migration and NAT rebinding).

```suggestion
1. Client and proxy exchange connection IDs via REGISTER_CLIENT_CID/REGISTER_TARGET_CID/ACK_CLIENT_CID/ACK_TARGET_CID and start using the forwarded mode.
1. Client probes a new network path and migrates.
1. Proxy, by receiving a non-probing packet on a non-active path that uses a Connection ID different from the active path, detects that the client intentionally migrated, and removes the forwarding rules.
1. Client re-registers the original client-to-target Connection IDs. This solicits new virtual CIDs from the proxy and does not count as an additional sequence number towards MAX_CONNECTION_IDS.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+1. Upon client-proxy network path validation, forwarding rules are configured with the new
+virtual CIDs provided by the proxy.
```
kazuho: I concur that the proxy should refrain from forwarding packets to the client until the path is validated, because until then, the amplification limit matters and because QUIC version 1 packets need to be sent to validate the path.

However, if the proxy sends an ACK_TARGET_CID frame but withholds from installing the forwarding rules, the client might send a packet with the newly advertised target VCID and that packet might get dropped by the proxy. This would have negative impact on the congestion control on the client-to-target connection.

Therefore, I think we should state either
* that a proxy withholds ACK_TARGET_CID/ACK_CLIENT_CID capsules until the path is validated, or
* that a proxy sends ACK_TARGET_CID and enables forwarding in the client-to-target direction immediately, but that the emission of ACK_CLIENT_CID and the enablement of the forwarding rule in the target-to-client direction is delayed until the new network path is validated.

Since we require forwarding rule in the client-to-target direction to be kept alive during path validation for the NAT-rebinding case, the suggested text below mirrors that:

```suggestion
1. When the proxy receives a REGISTER_TARGET_CID capsule, the proxy responds with an ACK_TARGET_CID capsule carrying a fresh VCID, and enables the forwarding rule in the client-to-target direction.
1. When the proxy receives a REGISTER_CLIENT_CID capsule, the proxy withholds a response until the new network path is validated. Once the new network path is validated, the proxy sends an ACK_CLIENT_CID capsule carrying a fresh VCID; once the client acknowledges it with an ACK_CLIENT_VCID capsule, the proxy enables the forwarding rule in the target-to-client direction.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
 
 When operating in forwarded mode, the proxy reconfigures or removes forwarding
 rules as the network path between the client and proxy changes. In the event of
-passive migration, the proxy automatically reconfigures forwarding rules to use
+passive migration, the proxy MUST automatically reconfigure forwarding rules to use
 the latest active and validated network path for the HTTP stream. In the event of
```
kazuho: Split paragraph.

Separately, I think we might adjust the change so that it cannot be interpreted that a proxy is required to drop packets in the direction of client-to-target, until the path is validated. It could be painful, since it would mean that the packets sent from the client after quiescence will be dropped if NAT rebinding happened, causing delays of multiple round-trips to recover.

```suggestion
passive migration, the proxy automatically reconfigures forwarding rules to use
the latest active and validated network path for the HTTP stream. Specifically,
the proxy MUST NOT forward packets to the client until the new network path
is validated. The proxy MAY continue forwarding packets to the target.

In the event of
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
-active migration, the proxy removes forwarding rules in order to not send
+active migration, the proxy MUST remove forwarding rules in order to not send
 packets with the same connection ID bytes over multiple network paths. After
```
kazuho:
```suggestion
packets with the same connection ID bytes over multiple network paths, and MUST NOT apply the new forwarding rule in the target-to-client direction until the new network path is validated. After
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+ID bytes being used over multiple network paths. Note that the client CID and target
+CID can stay the same while the target VCID and client VCID change. Importantly,
+the client and proxy do not send a CLOSE_CLIENT_CID capsule because that would also
+remove the registration of the CID for the purpose of port sharing, potentially breaking the
+tunnelled path.
+
+### Passive Migration Steps {#passive-migration-steps}
+
+1. Client registers connection IDs via REGISTER_CLIENT_CID/REGISTER_TARGET_CID
+1. Client experiences NAT rebinding
```
kazuho: This is a matter of taste, but the client experiences nothing. It is either that a network path experiences a NAT rebinding or that a proxy observes NAT rebinding.
```suggestion
1. The network path experiences NAT rebinding.
```

draft-ietf-masque-quic-proxy.md
```diff
@@ -1278,17 +1278,47 @@ path to avoid creating unintentional congestion on the new path.
...
+1. Proxy reconfigures forwarding rules to reuse the same virtual CID on the new network path.
+Forwarding rules activated upon validation of the client-proxy network path.
```
kazuho: Regarding why this should better be unidirectional; please see my comment above.

```suggestion
1. Proxy reconfigures the forwarding rule in the target-to-client direction to reuse the same virtual CID on the new network path.
```

draft-ietf-masque-quic-proxy.md
kazuho: Current text is inconsistent with RFC 9000 that sets the limit to the amplification factor.

Since this PR forbids forwarding in the proxy-to-client direction until the client's address is validated, we can simply say so. As for the tunnel mode, nothing needs to be stated in this specification, as datagrams are forwarded using the mechanism defined in RFC 9298.

```suggestion
Specifically, the proxy MUST NOT forward packets
to an unvalidated client address.
```

kazuho (commented):
Thank you for writing all of this @ehaydenr.

With all the description it is much much easier to understand and I think the direction is correct. Please see my comments inline.

DavidSchinazi (approved):
This is a good clarification. +1 to most of Kazuho's comments

draft-ietf-masque-quic-proxy.md
ehaydenr: Thank you for catching this - it definitely is contradictory. I'm a bit conflicted here.

Going with the original text prior to this PR where all registrations (initial and re-registration) consume a sequence number is convenient from a counting perspective and is the most cautious approach. The proxy is able to limit not just concurrent registrations, but also registration churn. Registration churn might be expensive depending on your implementation. The downside is the proxy has to be careful about issuance of MAX_CONNECTION_IDs in cases like migration or rejected registration so that the client always has spare credit to continue to register as needed. There's also the risk of naive implementations that get into a loop like you called out [here](https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/issues/160).

If we move towards the text in this PR, where re-registrations do not consume sequence numbers, we maintain limits around concurrent registrations, but we no longer bound registration churn. Migrations are simpler because you never need to think about MAX_CONNECTION_IDs since it's also just re-registration.

I'm partial to the original text prior to this PR which treats registrations more uniformly against MAX_CONNECTION_IDs because it pushes proxies to think about registration churn limits and gives a non-fatal option for applying registration back pressure. In this case, proxies that inhibit clients from doing re-registrations are effectively forcing those clients back into tunneled mode. In the "re-registrations don't consume" mode, proxies that want to slow down or inhibit re-registration have no mechanism other than resetting the stream.

The impression I'm getting from your comments is you may be leaning the other way - towards re-registrations not consuming a sequence number. Is that right?

draft-ietf-masque-quic-proxy.md
ehaydenr: For "Proxy, by receiving ...", should that be a "non-active, validated path"? Otherwise, this looks good to me, but we should sort out https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/pull/170#discussion_r3525538787 first.

draft-ietf-masque-quic-proxy.md
ehaydenr: Why delay emission of the ACK_CLIENT_CID? Why can't the proxy send that immediately, allowing port sharing right away through tunneled mode, and then after both receiving ACK_CLIENT_VCID and the path being validated that the forwarding rule be enabled?

draft-ietf-masque-quic-proxy.md
ehaydenr: I think this makes sense. I'm wondering if we should change from `MAY` to `SHOULD` here - what do you think?

draft-ietf-masque-quic-proxy.md
ehaydenr: Upon NAT rebinding, I would think:
1. proxy immediately and automatically reconfigures client->target forwarding rules to prevent dropping of forwarded mode packets.
2. proxy waits for client-proxy network path to become validated and then automatically reconfigures target->client forwarding rules.

Is that what you're thinking as well or are you suggesting something different?

draft-ietf-masque-quic-proxy.md
kazuho: I share the view that consuming a sequence number is better (sorry if I confused you).

PS. Though, if multiple REGISTER_*_CID capsules carrying the same CID are to be assigned different sequence numbers, wouldn't those entries no longer be individually closable, since the close capsules identify entries by CID rather than sequence number?

draft-ietf-masque-quic-proxy.md
kazuho: Agreed. SHOULD is the right call, because not forwarding means drop in the client-to-target direction.

draft-ietf-masque-quic-proxy.md
kazuho: > Upon NAT rebinding, I would think:
> 1. proxy immediately and automatically reconfigures client->target forwarding rules to prevent dropping of forwarded mode packets.
> 1. proxy waits for client-proxy network path to become validated and then automatically reconfigures target->client forwarding rules.

This works for me.

Honestly, I'm not sure if there is anything that needs to be reconfigured in the client-to-target direction, as the proxy will recognize the client-sent forwarded-mode packets they receive from _any_ address (that is the premise of handling NAT rebinding). Though that's a minor detail.

draft-ietf-masque-quic-proxy.md
kazuho: > For "Proxy, by receiving ...", should that be a "non-active, validated path"?

RFC 9000 doesn't assume the new path is validated. Pre-migration probing is a MAY ([Section 9.1](https://datatracker.ietf.org/doc/html/rfc9000#name-probing-a-new-path)), so a proxy can legitimately detect active migration onto a not-yet-validated path. That's the normal case whenever the client can't seamlessly pre-validate: e.g., a device with no cellular that drops one Wi-Fi and comes up on another.

If the new path happens to already be validated, the proxy is in luck and can collapse the steps here (enable target-to-client forwarding immediately). But that's the fortunate case; we need to write down the correct handling for the unvalidated case — so the detection step shouldn't be conditioned on "validated," and the validation gate stays on re-enabling target-to-client forwarding.

draft-ietf-masque-quic-proxy.md
kazuho: Good point. I think that works.

draft-ietf-masque-quic-proxy.md
ehaydenr: This is a good question. Consider:
1. Client registers
2. Proxy acknowledges
3. At the same time, client re-registers and proxy closes original registration
4. Proxy receives client's re-registration, but interprets as fresh registration
5. Client receives proxy close of original registration but treats it as a close (rejection) of the re-registration
6. Proxy thinks registration is active, but client thinks it is closed.

tfpauly: Suggestion:
- Agree that re-register should consume a sequence number
- Split close capsules into "reject" (that only proxy can send, and can only send in response to the request, and not after having acknowledged -- so a proxy either accepts or rejects, and cannot close after accept); and "close" that only the client can send
- Require proxies to respond to all registration requests in the order in which the client sent them; each needs an explicit response (reject or ack)

ehaydenr: Implemented the suggestion in https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/pull/172

draft-ietf-masque-quic-proxy.md
ehaydenr: clarified in 0ed6fc6f82f7e6dd8107fa2a4fa5e85de982592d

draft-ietf-masque-quic-proxy.md
ehaydenr: f6b4ad44f934fede4c6035a9399f74297b92bd4e

PR #172, opened 2026-07-06
split out close and rejection

ehaydenr:
https://github.com/ietf-wg-masque/draft-ietf-masque-quic-proxy/pull/170#discussion_r3532255895

### DIFF
--- a/draft-ietf-masque-quic-proxy.md
+++ b/draft-ietf-masque-quic-proxy.md
@@ -495,23 +495,22 @@ acknowledgement. This capsule also contains a Stateless Reset Token the client
 may respond with when receiving forwarded mode packets with the specified
 virtual connection ID.
 
+The REJECT_CLIENT_CID and REJECT_TARGET_CID capsule types ({{capsule-reject}})
+are used to reject connection ID registrations. Each capsule includes
+a reason code indicating the reason for the rejection. These capsules MUST only be sent
+by the proxy. A proxy MUST NOT send a rejection capsule for a registration that it
+has already acknowledged. If a client receives such a capsule, it MUST reset the
+stream with H3_DATAGRAM_ERROR error code.
+
 The CLOSE_CLIENT_CID and CLOSE_TARGET_CID capsule types ({{capsule-close}})
-are used to close or reject connection ID registrations. Each capsule includes
-a reason code indicating why the connection ID is being closed.
+are used to close connection ID registrations. These capsules are only sent by the
+client and MUST NOT be sent by proxies.
 
 Clients send CLOSE_CLIENT_CID or CLOSE_TARGET_CID capsules to retire connection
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

draft-ietf-masque-quic-proxy.md
```diff
@@ -495,24 +495,23 @@ acknowledgement. This capsule also contains a Stateless Reset Token the client
 may respond with when receiving forwarded mode packets with the specified
 virtual connection ID.
 
+The REJECT_CLIENT_CID and REJECT_TARGET_CID capsule types ({{capsule-reject}})
+are used to reject connection ID registrations. Each capsule includes
+a reason code indicating the reason for the rejection. These capsules MUST only be sent
+by the proxy. A proxy MUST NOT send a rejection capsule for a registration that it
+has already acknowledged. If a client receives such a capsule, it MUST reset the
```
kazuho, line 502: IIUC, this restriction means that a proxy is no longer able to retire a CID that it issued without breaking the connection. Nor does the proxy have a way to ask the client to stop using forwarding mode in the client-to-target direction.

Assuming that my understanding is correct, I'm concerned that such a restriction may not align with the rules governing CIDs in RFC 9000, nor be compatible with [QUIC-LB](https://datatracker.ietf.org/doc/draft-ietf-quic-load-balancers/).

RFC 9000 allows a QUIC server (i.e., the masque proxy) to retire CIDs it issued. This is a building block for QUIC servers that encrypt the CIDs they issue using a shared key that is rotated. Quoting from the RFC and the I-D:

> An endpoint might need to stop accepting previously issued connection IDs in certain circumstances. Such an endpoint can cause its peer to retire connection IDs by sending a NEW_CONNECTION_ID frame with an increased Retire Prior To field. ([Section 5.1.2 of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-5.1.2-4))

> Servers MUST use that QUIC version's methods to update the client with CIDs (e.g., NEW_CONNECTION_ID frames) using the new configuration and retire CIDs using the old configuration. ([Section 3.2 of draft-ietf-quic-load-balancers-21](https://www.ietf.org/archive/id/draft-ietf-quic-load-balancers-21.html#section-3.1-5))

Aside from those that implement QUIC-LB, there are QUIC servers that use a proprietary encryption scheme with key rotation (we do).

Considering that the forwarding mode is an optimization, I do not think we need something like a Retire Prior To field; all that is needed is a mechanism that allows a proxy to either a) ask the client to re-run the CID registration and obtain a new target VCID, or b) ask the client to stop using forwarding mode.

However, as specified, forwarding mode provides neither (a) nor (b): a proxy has no way to retire a target VCID after acknowledging it, and no fallback signal is defined. Forwarding mode now lacks a capability that RFC 9000 provides, making it unusable for proxies that run the kind of key-rotating CID scheme described above.

